PR #20, opened 2022-02-15
Add remote upstream step in contributing steps

christofid:


drugilsberg (approved):
Looks great thanks for catching this.

PR #33, opened 2022-02-18
Black release version

C-nit:
Updating the version of black in the dev_requirements.txt, as well as a run of black formatting the code.

This PR is also to test the workflow to sign the CLA, and see if there is an effect when not signing off git commits explicitly.


drugilsberg (approved):
LGTM good job, I was expecting more changes :)

PR #36, opened 2022-02-21
Exposed paccmann based BiModal MCA affinity predictor

YoelShoshan:
* Exposed paccmann based BiModal MCA affinity predictor
* Added a regression unit test for example case
* NOTE - verified on the current main pip version of pytoda, will not work with latest pytoda due to Pacmmann_GP language files being older version
* NOTE - requires part of paccmann_gp resources files in order to work well (to test I manually copied resources to ~/.gt4sd/... to verify it works)


drugilsberg (commented):
Hey @YoelShoshan the work looks great, well done, I will suggest a couple of changes and a refactor, but it's truly a nice piece of work 👏

drugilsberg (approved):

PR #42, opened 2022-02-22
Improve CLA workflow

C-nit:
actions to commit to other peoples forks was not something super easy to do, so I'm settling for a bit more verbosity and automation.

the issue will be closed with a comment to the commit that added the contributor. There is a notice to merge this into a PR.

Therefore there is no assignment of the issue any more.

Looks like this: https://github.com/C-nit/gt4sd-core/issues/9
and can also be triggered in a different way: https://github.com/C-nit/gt4sd-core/issues/11


C-nit:
cla bot is only blocking because there are commits by the fictional 
```
          git config --local user.email "gt4sd-actions@github.com"
          git config --local user.name "GitHub Action"
```

C-nit:
This one should really be a squash.

drugilsberg:
> 

do you have any idea how to fix this? maybe adding fictional user to contributors? or set one of the repo maintainers as commiter?

drugilsberg (approved):
Looks great, see only the comment on the fictional user.

C-nit:
> do you have any idea how to fix this? maybe adding fictional user to contributors? or set one of the repo maintainers as commiter?

I think it's a not an issue that will arise again. Here I was messing on the main on my fork to test the actions, and some of it's commits made it into the PR.

However in the usual flow Actions will commit to master. The merge commit of that in the PR is not a problem for the bot.

So I would just force the merge now. We have to revisit this only when we have GH Actions committing to PRs.

C-nit:
@drugilsberg before I merge one thing that is hidden by the diff of the indentation:
I changed to use `secrets.GITHUB_TOKEN` that is an automatic ephemeral token set up per run.

I wanted to confirm there was no specific reason to use our own long lived token?

drugilsberg:
> en by the diff of the indentation:
> I changed to use `secrets.GITHUB_TOKEN` that is an

totally ok with it, I wanted to operate the change anyhow, could you also use it in the Sphinx docs generation pipeline?

C-nit:
Done, I also restricted permissions of the GITHUB_TOKEN with a bit of guessing what scopes are needed. Unfortunately the scopes itself are not well documented.

drugilsberg (approved):
LGTM, feel free to squash and merge

PR #46, opened 2022-02-25
Add PGT algorithm

christofid:


drugilsberg (approved):
Looks great to me. Thanks for the amazing work. I would just wait for the CI to pass after having synced with main and then you can squash and merge.

PR #49, opened 2022-03-10
Torchdrug integration

jannisborn:
TorchDrug integration (inference-only) as discussed @drugilsberg!

- 2 models implemented `GCPN` (Graph Convolutional Policy Network, NeurIPS 2018) and `GAF` (Graph AutoregressiveFlow, ICLR 2020).
- For each models, 3 pretrained models are made available: 
  - zinc250k: `TorchDrugZincGCPN` and `TorchDrugZincGAF`
  - optimized to generated high-qed molecules: `TorchDrugQedGAF` and `TorchDrugQedGCPN`.
  - optimized to generated high-plogp molecules: `TorchDrugPlogpGCPN` and `TorchDrugPlogpGAF`

- unittests implemented for all models

ToDo:
- [x] fix CI pipeline
- [x] upload models on COS (2 times)

## Regarding CI: 

I knew this would backfire, the tests are still failing at installation. Locally, I had to install `pytorch-scatter` via conda, I did the following:
```sh
pip3 install torch==1.8.1 torchvision==0.9.1 torchaudio==0.8.1 -f https://download.pytorch.org/whl/lts/1.8/torch_lts.html
conda install pytorch-scatter -c pyg
pip install torchdrug
```
and this worked fine.  

When I reproduce the CI pipeline locally, I also do not get a functioning installation. The env is created successfully but importing `pytorch-scatter` results in segmentation faults.

Possible solution: Adapt the `conda.yml` like so:
```yml
name: gt4sd
channels:
  - https://conda.anaconda.org/pyg
dependencies:
  - python>=3.7,<3.8
  - pip>=19.1,<20.3
  - pytorch-scatter
  - pip:
      - -r requirements.txt
      - -r vcs_requirements.txt
      # development
      - -r dev_requirements.txt
```

### DIFF
--- a/conda.yml
+++ b/conda.yml
@@ -5,11 +5,10 @@ channels:
 dependencies:
   - python>=3.7,<3.8
   - pip>=19.1,<20.3
-  - pytorch
-  - pytorch-scatter
+  - pytorch>=1.0
+  - pytorch-scatter>=2.0.0
   - pip:
       - -r requirements.txt
       - -r vcs_requirements.txt
       # development
       - -r dev_requirements.txt
-  - pytorch-scatter
--- a/conda.yml
+++ b/conda.yml
@@ -1,12 +1,15 @@
 name: gt4sd
 channels:
   - https://conda.anaconda.org/pyg
+  - https://conda.anaconda.org/pytorch
 dependencies:
   - python>=3.7,<3.8
   - pip>=19.1,<20.3
+  - pytorch
   - pytorch-scatter
   - pip:
       - -r requirements.txt
       - -r vcs_requirements.txt
       # development
       - -r dev_requirements.txt
+  - pytorch-scatter
--- a/conda.yml
+++ b/conda.yml
@@ -1,7 +1,10 @@
 name: gt4sd
+channels:
+  - https://conda.anaconda.org/pyg
 dependencies:
   - python>=3.7,<3.8
   - pip>=19.1,<20.3
+  - pytorch-scatter
   - pip:
       - -r requirements.txt
       - -r vcs_requirements.txt
--- a/src/gt4sd/algorithms/generation/torchdrug/abc.py
+++ b/src/gt4sd/algorithms/generation/torchdrug/abc.py
@@ -1,4 +1,6 @@
 import torch
 
 # Disable openmp usage since this raises on MacOS when libomp has the wrong version.
-torch._C.has_openmp = False
+openmp = torch._C.has_openmp
+if openmp:
+    torch._C.has_openmp = False
--- a/src/gt4sd/algorithms/generation/torchdrug/abc.py
+++ b/src/gt4sd/algorithms/generation/torchdrug/abc.py
@@ -1,0 +1,4 @@
+import torch
+
+# Disable openmp usage since this raises on MacOS when libomp has the wrong version.
+torch._C.has_openmp = False
--- a/src/gt4sd/algorithms/generation/torchdrug/abc.py
+++ b/src/gt4sd/algorithms/generation/torchdrug/abc.py
@@ -1,6 +1,6 @@
 import torch
 
-# Disable openmp usage since this raises on MacOS when libomp has the wrong version.
+# disable openmp usage since this raises on MacOS when libomp has the wrong version.
 openmp = torch._C.has_openmp
 if openmp:
     torch._C.has_openmp = False
--- a/src/gt4sd/algorithms/generation/torchdrug/core.py
+++ b/src/gt4sd/algorithms/generation/torchdrug/core.py
@@ -21,14 +21,15 @@ class TorchDrugGenerator(GeneratorAlgorithm[S, T]):
         """TorchDrug generation algorithm.
 
         Args:
-            configuration: domain and application
-                specification, defining types and validations.
+            configuration: domain and application specification, defining types
+                and validations.  Currently supported algorithm versions are:
+                "zinc250k_v0", "qed_v0" and "plogp_v0".
             target: unused since it is not a conditional generator.
 
         Example:
             An example for using a generative algorithm from HuggingFace::
 
-                configuration = TorchDrugZincGCPN()
+                configuration = TorchDrugGCPN(algorithm_version="qed_v0")
                 algorithm = TorchDrugGenerator(configuration=configuration)
                 items = list(algorithm.sample(1))
                 print(items)
@@ -69,10 +70,15 @@ class TorchDrugGenerator(GeneratorAlgorithm[S, T]):
 
 
 @ApplicationsRegistry.register_algorithm_application(TorchDrugGenerator)
-class TorchDrugGCPNConfiguration(AlgorithmConfiguration[str, None]):
+class TorchDrugGCPN(AlgorithmConfiguration[str, None]):
+    """
+    Interface for TorchDrug Graph-convolutional policy network (GCPN) algorithm.
+    Currently supported algorithm versions are "zinc250k_v0", "qed_v0" and "plogp_v0".
+    """
 
     algorithm_type: ClassVar[str] = "generation"
     domain: ClassVar[str] = "materials"
+    algorithm_version: str = "zinc250k_v0"
 
     def get_conditional_generator(self, resources_path: str) -> GCPNGenerator:
         """Instantiate the actual generator implementation.
@@ -86,25 +92,15 @@ class TorchDrugGCPNConfiguration(AlgorithmConfiguration[str, None]):
 
 
 @ApplicationsRegistry.register_algorithm_application(TorchDrugGenerator)
-class TorchDrugZincGCPN(TorchDrugGCPNConfiguration):
-    algorithm_version: str = "v0"
-
-
-@ApplicationsRegistry.register_algorithm_application(TorchDrugGenerator)
-class TorchDrugQedGCPN(TorchDrugGCPNConfiguration):
-    algorithm_version: str = "v0"
-
-
-@ApplicationsRegistry.register_algorithm_application(TorchDrugGenerator)
-class TorchDrugPlogpGCPN(TorchDrugGCPNConfiguration):
-    algorithm_version: str = "v0"
-
-
-@ApplicationsRegistry.register_algorithm_application(TorchDrugGenerator)
-class TorchDrugGAFConfiguration(AlgorithmConfiguration[str, None]):
+class TorchDrugGraphAF(AlgorithmConfiguration[str, None]):
+    """
+    Interface for TorchDrug flow-based autoregressive graph algorithm (GraphAF).
+    Currently supported algorithm versions are "zinc250k_v0", "qed_v0" and "plogp_v0".
+    """
 
     algorithm_type: ClassVar[str] = "generation"
     domain: ClassVar[str] = "materials"
+    algorithm_version: str = "zinc250k_v0"
 
     def get_conditional_generator(self, resources_path: str) -> GAFGenerator:
         """Instantiate the actual generator implementation.
@@ -115,18 +111,3 @@ class TorchDrugGAFConfiguration(AlgorithmConfiguration[str, None]):
         """
         self.generator = GAFGenerator(resources_path=resources_path)
         return self.generator
-
-
-@ApplicationsRegistry.register_algorithm_application(TorchDrugGenerator)
-class TorchDrugZincGAF(TorchDrugGAFConfiguration):
-    algorithm_version: str = "v0"
-
-
-@ApplicationsRegistry.register_algorithm_application(TorchDrugGenerator)
-class TorchDrugQedGAF(TorchDrugGAFConfiguration):
-    algorithm_version: str = "v0"
-
-
-@ApplicationsRegistry.register_algorithm_application(TorchDrugGenerator)
-class TorchDrugPlogpGAF(TorchDrugGAFConfiguration):
-    algorithm_version: str = "v0"
--- a/src/gt4sd/algorithms/generation/torchdrug/core.py
+++ b/src/gt4sd/algorithms/generation/torchdrug/core.py
@@ -104,7 +104,7 @@ class TorchDrugPlogpGCPN(TorchDrugGCPNConfiguration):
 class TorchDrugGAFConfiguration(AlgorithmConfiguration[str, None]):
 
     algorithm_type: ClassVar[str] = "generation"
-    domain: ClassVar[str] = "material"
+    domain: ClassVar[str] = "materials"
 
     def get_conditional_generator(self, resources_path: str) -> GAFGenerator:
         """Instantiate the actual generator implementation.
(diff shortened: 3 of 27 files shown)


jannisborn:
Open for review and discussion about pytorch-scatter/torchdrug installation workflow @drugilsberg!

jannisborn:
Final updates on this PR @drugilsberg:
- I parametrized `num_sample` for the generation, as we discussed. I checked it locally, the sample generation is faster now
- openmp is enabled again after the torchdrug imports and it does not interfer with torchdrug-related inference, even if we mix it with inference of other models
- `pytorch-scatter` installation: the package installs stable on ubuntu if we install pytorch via conda too. I changed the `conda.yml`, and the installations runs through now, no more segmentation faults on both mac and ubuntu 🚀  The conda incubator setup goes up from 4 to 10min but I guess that's the price we have to pay for torchdrug 😐 
- I found (and fixed) a terrible bug caused by `torchdrug`. Essentially, they overwrite the default `nn.Module` of torch. See [here](https://github.com/DeepGraphLearning/torchdrug/blob/eeee19181572ef5b8a806b71bdd4d2d1a4e27f67/torchdrug/patch.py#L125). The crazy thing is that **all** subsequent code running native torch uses this patched module. I found it out because they forgot to implement a keyword argument that is present in native torch and that was used by the EnzymeOptimzer (Roberta). I patched it in `torchdrug.implementation`, see bda7951d96ec9573761d2d8999155969d57f5df3 but overall the torchdrug integration raises some concerns regarding the overall code reliability 😐  I opened an issue in the torchdrug repo, but I think we can merge this anyways.

conda.yml
```diff
@@ -1,9 +1,15 @@
...
   - python>=3.7,<3.8
   - pip>=19.1,<20.3
+  - pytorch
+  - pytorch-scatter
   - pip:
       - -r requirements.txt
       - -r vcs_requirements.txt
       # development
       - -r dev_requirements.txt
+  - pytorch-scatter
```
drugilsberg, line 15: this should probably be removed

conda.yml
```diff
@@ -1,9 +1,15 @@
 name: gt4sd
+channels:
+  - https://conda.anaconda.org/pyg
+  - https://conda.anaconda.org/pytorch
 dependencies:
   - python>=3.7,<3.8
   - pip>=19.1,<20.3
+  - pytorch
```
drugilsberg, line 8: not a fan but if we have to keep it I would provide a lower bound like we do in requirements.txt

src/gt4sd/algorithms/generation/torchdrug/implementation.py
```diff
@@ -0,0 +1,192 @@
...
+    task: GCPNGeneration
+
+    input_dim: int = 18
+    num_relation: int = 3
+    batch_norm: bool = False
+    atom_types: List[int] = [6, 7, 8, 9, 15, 16, 17, 35, 53]
+    hidden_dims: List[int] = [256, 256, 256, 256]
+
+    def __init__(self, resources_path: str):
+        """
```
drugilsberg, line 121: add initial comment in docstring

src/gt4sd/frameworks/granular/arg_parser/parser.py
```diff
@@ -94,3 +98,6 @@ def parse_arguments_from_config(conf_file: Optional[str] = None) -> argparse.Nam
     result_namespace = argparse.Namespace(**result)
 
     return result_namespace
+
+
+_sentencepiece
```
drugilsberg, line 103: typo? I would remove this

src/gt4sd/algorithms/generation/torchdrug/core.py
```diff
@@ -0,0 +1,132 @@
...
+            resources_path: local path to model files.
+        Returns:
+            instance with :meth:`sample<gt4sd.algorithms.generation.torchdrug.implementation.GCPNGenerator.sample>` method for generation.
+        """
+        self.generator = GCPNGenerator(resources_path=resources_path)
+        return self.generator
+
+
+@ApplicationsRegistry.register_algorithm_application(TorchDrugGenerator)
+class TorchDrugZincGCPN(TorchDrugGCPNConfiguration):
```
drugilsberg, line 89: Here I think we should have a unique configuration per network type.
No need to create one for each dataset. We can use the algorithm version to treat all datasets.
We can have as default `zinc250k` and then the others are simply other folders in the cache.
The `ApplicationRegistry` will take care of everything.

drugilsberg (changes requested):
Great job some general comments we need to address before merging:

- I would not have configuration classes for specific datasets. Since the classes are the same we should simply use versions for the different datatsets (see comment in PR).
- `sentencepiece` treatment. Instead of doing the trick every time I would do it once in  `src/gt4sd/__init__.py`. After the definition of `__version__` and `__name__` we can have the following:
```python
import sentencepiece as _sentencepiece  # noqa: F401
import pytorch_lightning as _pl  # noqa: F401
```
In this way we only have it there and inside the module we can import pytorch_lightning without any issue.
- I left a note on the installation.

Once those are addressed I would say we are good to go.

conda.yml
jannisborn: Done!

conda.yml
jannisborn: not a fan at all either. If you have a better solution, I'd be more than happy to implement it.

jannisborn: I set lower bounds to pytorch and pytorch-scatter now

src/gt4sd/algorithms/generation/torchdrug/implementation.py
jannisborn: Done in [04fd6f6](https://github.com/GT4SD/gt4sd-core/pull/49/commits/04fd6f6d79c2703828c5c61c935673c2726eefad)

src/gt4sd/frameworks/granular/arg_parser/parser.py
jannisborn: It's not a typo. This is needed because flake8 complains otherwise. The tests are executed with the `--disable-noqa` flag, so I dont see any other way of circumvening flake8 from complaining. The current configuration only ignores `__init__` files, see here: https://github.com/GT4SD/gt4sd-core/blob/d127590e2a312bda6c55f39304b70462f342ef4a/.github/workflows/tests.yaml#L45

If we import sentencepiece in the file, we have to use it somehow, otherwise we get a F401

src/gt4sd/algorithms/generation/torchdrug/core.py
jannisborn: Refactored this in [13a843f](https://github.com/GT4SD/gt4sd-core/pull/49/commits/13a843fd1622d5488eb76b895fd6a9501798ed2d), it's definitely more concise now

jannisborn:
I addressed all comments @drugilsberg, but unfortunately your solution with importing `sentencepiece` in the global init file does not work. Dependent on which functions we import first from the package, the `__init__` is not always executed before an import of lightning.

src/gt4sd/algorithms/generation/torchdrug/abc.py
```diff
@@ -0,0 +1,4 @@
+import torch
```
drugilsberg, line 1: Do we still need this after the change in the setup? If yes, we should find to find a way to renable OpenMP after the import as quickly discussed.

src/gt4sd/algorithms/generation/torchdrug/core.py
```diff
@@ -0,0 +1,113 @@
...
+        """TorchDrug generation algorithm.
+
+        Args:
+            configuration: domain and application specification, defining types
+                and validations.  Currently supported algorithm versions are:
+                "zinc250k_v0", "qed_v0" and "plogp_v0".
+            target: unused since it is not a conditional generator.
+
+        Example:
+            An example for using a generative algorithm from HuggingFace::
```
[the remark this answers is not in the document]
jannisborn: Good catch, resolved this. I did a grep, there's no other occurrences of this

src/gt4sd/algorithms/generation/torchdrug/abc.py
jannisborn: Yes, we still need this and this setup is indeed a way to renable openMP after the torchdrug import. 
We disable openMP in `abc.py` which always runs first in the init. In `core.py` we import the torchdrug things and then afterwards we disable it here again.

src/gt4sd/frameworks/granular/arg_parser/parser.py
jannisborn: As I wrote above, this will not work because we run flake8 with `--disable-noqa` so we cannot overrule F401 errors, unless in `init` files.

src/gt4sd/algorithms/generation/torchdrug/abc.py
jannisborn: Same problem like for the sentencepiece issue, if I put the the line to re-enable openmp in the torchdrug file, directly after the last torchdrug import, we get a E402 since we dont have all module imports at the top

jannisborn: We could move it below the last import in the `implementation.py` but then I find it cleaner here especially since this init file is always executed even if we start a shell like:
```py
from gt4sd.algorithms.generation.torchdrug.implementation import *
```

drugilsberg: makes sense 👍

src/gt4sd/frameworks/granular/arg_parser/parser.py
drugilsberg: I would honestly prefer to see _sentencepiece just below the import because it's a bit weird to see this there without comments.

drugilsberg (commented):
Looks good great job, just few comments remained.

src/gt4sd/algorithms/generation/torchdrug/abc.py
drugilsberg: Actually we could simply put a check and set it to False, and optionally reenable it, only in case had_openmp is indeed True at the time of import. Otherwise we risk to enable it even if OpenMP is not there.

jannisborn: Done in 694b57f

src/gt4sd/training_pipelines/pytorch_lightning/core.py
```diff
@@ -4,13 +4,17 @@
...
 
+import sentencepiece as _sentencepiece
 from pytorch_lightning import LightningDataModule, LightningModule, Trainer
 from pytorch_lightning.callbacks.base import Callback
 from pytorch_lightning.callbacks.early_stopping import EarlyStopping
 from pytorch_lightning.callbacks.model_checkpoint import ModelCheckpoint
 
 from ..core import TrainingPipeline, TrainingPipelineArguments
 
+# Sentencepiece has to be loaded before lightning
```
drugilsberg, line 15:
```suggestion
# sentencepiece has to be loaded before lightning to avoid segfaults
```

src/gt4sd/training_pipelines/pytorch_lightning/language_modeling/core.py
```diff
@@ -4,13 +4,17 @@
...
 
+import sentencepiece as _sentencepiece
 from pytorch_lightning import LightningDataModule, LightningModule
 
 from ...core import TrainingPipelineArguments
 from ..core import PyTorchLightningTrainingPipeline
 from .lm_datasets import CGMDataModule, CLMDataModule, MLMDataModule, PLMDataModule
 from .models import LM_MODULE_FACTORY, CGMModule, CLMModule, MLMModule, PLMModule
 
+# Sentencepiece has to be loaded before lightning
```
drugilsberg, line 15:
```suggestion
# sentencepiece has to be loaded before lightning to avoid segfaults
```

src/gt4sd/training_pipelines/pytorch_lightning/language_modeling/models.py
```diff
@@ -19,6 +20,9 @@
     XLNetLMHeadModel,
 )
 
+# Sentencepiece has to be loaded before lightning
```
drugilsberg, line 23:
```suggestion
# sentencepiece has to be loaded before lightning to avoid segfaults
```

src/gt4sd/frameworks/granular/arg_parser/parser.py
```diff
@@ -2,11 +2,15 @@
 import configparser
 from typing import Any, Dict, Optional
 
+import sentencepiece as _sentencepiece
 from pytorch_lightning import Trainer
 
 from ..ml.models import ARCHITECTURE_FACTORY
 from .utils import convert_string_to_class
 
+# Sentencepiece has to be loaded before lightning
```
drugilsberg, line 11:
```suggestion
# sentencepiece has to be loaded before lightning to avoid segfaults
```

src/gt4sd/frameworks/granular/dataloader/data_module.py
```diff
@@ -11,6 +12,9 @@
 from .dataset import CombinedGranularDataset, GranularDataset
 from .sampler import StratifiedSampler
 
+# Sentencepiece has to be loaded before lightning
```
drugilsberg, line 15:
```suggestion
# sentencepiece has to be loaded before lightning to avoid segfaults
```

src/gt4sd/frameworks/granular/ml/module.py
```diff
@@ -3,13 +3,17 @@
...
 
+import sentencepiece as _sentencepiece
 import pandas as pd
 import pytorch_lightning as pl
 import torch
 
 from .models import GranularBaseModel, GranularEncoderDecoderModel
 from .models.model_builder import building_models, define_latent_models_input_size
 
+# Sentencepiece has to be loaded before lightning
```
drugilsberg, line 14:
```suggestion
# sentencepiece has to be loaded before lightning to avoid segfaults
```

src/gt4sd/frameworks/granular/train/core.py
```diff
@@ -14,6 +15,9 @@
 from ..ml.models import AUTOENCODER_ARCHITECTURES
 from ..ml.module import GranularModule
 
+# Sentencepiece has to be loaded before lightning
```
drugilsberg, line 18:
```suggestion
# sentencepiece has to be loaded before lightning to avoid segfaults
```

src/gt4sd/algorithms/generation/torchdrug/abc.py
```diff
@@ -0,0 +1,6 @@
...
+
+# Disable openmp usage since this raises on MacOS when libomp has the wrong version.
```
drugilsberg, line 3:
```suggestion
# disable openmp usage since this raises on MacOS when libomp has the wrong version.
```

drugilsberg (approved):
Looks great well done

PR #50, opened 2022-03-11
Enable LM saving

christofid:
Enable the use of saving-cli for Language Models. The given language models could be: 
-  a hugging face model (including its tokenizer) or 
-  a pytorch-lightning checkpoint.
 
In the later case, a conversion to hugging face format will also take place during saving.

### DIFF
--- a/src/gt4sd/algorithms/core.py
+++ b/src/gt4sd/algorithms/core.py
@@ -439,11 +439,11 @@ class AlgorithmConfiguration(Generic[S, T]):
         )
 
     @classmethod
-    def save_postprocess(
+    def save_version_from_training_pipeline_arguments_postprocess(
         cls,
         training_pipeline_arguments: TrainingPipelineArguments,
     ):
-        """Postprocess after saving
+        """Postprocess after saving.
 
         Args:
             training_pipeline_arguments: training pipeline arguments.
@@ -506,7 +506,9 @@ class AlgorithmConfiguration(Generic[S, T]):
                 )
                 shutil.copyfile(source_filepath, target_filepath)
 
-            cls.save_postprocess(training_pipeline_arguments)
+            cls.save_version_from_training_pipeline_arguments_postprocess(
+                training_pipeline_arguments
+            )
 
             logger.info(f"Artifacts saving completed into {target_path}")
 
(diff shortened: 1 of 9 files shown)


src/gt4sd/algorithms/generation/pgt/core.py
```diff
@@ -107,10 +113,7 @@ class PGTAlgorithmConfiguration(AlgorithmConfiguration[str, None]):
     domain: ClassVar[str] = "nlp"
     algorithm_version: str = "v0"
 
-    prompt: str = field(
-        default="I am a interesting prompt",
-        metadata=dict(description="A prompt input for generation."),
-    )
+    prompt: str = "Do not change me, I am an interesting prompt that is changed internally for each task."
```
drugilsberg, line 116: do we need this define in the configuration?

src/gt4sd/algorithms/generation/pgt/core.py
```diff
@@ -168,6 +171,38 @@ def get_generator(self, resources_path: str, **kwargs) -> Generator:
...
+
+        Returns:
+            a mapping between artifacts' files and training pipeline's output files.
+        """
+
+        if isinstance(training_pipeline_arguments, LanguageModelingSavingArguments):
+
+            if training_pipeline_arguments.ckpt is not None:
+
+                convert_pl_to_hf(training_pipeline_arguments)
```
drugilsberg, line 191: I like it a lot but I do not like the fact we produce files that are then simply moved, but for the time being is ok

drugilsberg (approved):
Once checks pass we can merge

src/gt4sd/algorithms/core.py
```diff
@@ -438,6 +438,18 @@ def get_filepath_mappings_for_training_pipeline_arguments(
             f"{cls.__name__} artifacts not mapped for {training_pipeline_arguments}"
         )
 
+    @classmethod
+    def save_postprocess(
```
drugilsberg, line 442: I would rename to `def save_version_from_training_pipeline_arguments_postprocess`

src/gt4sd/algorithms/core.py
```diff
@@ -438,6 +438,18 @@ def get_filepath_mappings_for_training_pipeline_arguments(
...
+        cls,
+        training_pipeline_arguments: TrainingPipelineArguments,
+    ):
+        """Postprocess after saving
```
drugilsberg, line 446: remember the "."-s in docstrings.

drugilsberg (commented):
Small comment then all good to go

Issue #10, opened 2022-02-11
include support and interfaces for torchdrug


drugilsberg:

jannisborn:
- [ ] Inference pipelines (done in #49).
- [ ] training pipelines

PR #56, opened 2022-04-04
Torchdrug trainer pipeline

jannisborn:
Implemented torchdrug trainer pipeline.
Models can be used via:
```sh
gt4sd-trainer --training_pipeline_name torchdrug-gcpn-trainer -h
gt4sd-trainer --training_pipeline_name torchdrug-graphaf-trainer -h
```

Features: 
- [x] Support for the same two models that are available via inference `TorchDrugGCPN` and `TorchDrugGraphAF`.
- [x] Both models can be trained on all `MoleculeDatasets` from `torchdrug.Datasets`. Those are around 20 predefined datasets.
- [x] Implemented a custom dataset where users can pass their own data
- [x] In addition to the unittests I verified functionalities from the CLI via `gt4sd-trainer`

Problems:
- [ ] Property optimization does not work, due to instabilities in TorchDrug. I opened issue and PR but we have to wait until they merge, release a new version and then bump our dependency. The code I wrote here already supports the property optimization but I disabled the unittest for the moment because it would fail due to the TorchDrug issue. See details: https://github.com/DeepGraphLearning/torchdrug/issues/83
- [x] `gt4sd-saving`: I ran a test via CLI but the saving failed. Not sure how problematic this is, here's the error:
```sh
INFO:gt4sd.cli.saving:Selected configuration: ConfigurationTuple(algorithm_type='generation', domain='materials', algorithm_name='TorchDrugGenerator', algorithm_application='TorchDrugGCPN')
INFO:gt4sd.cli.saving:Saving model version "fast" with the following configuration: <class 'gt4sd.algorithms.generation.torchdrug.core.TorchDrugGCPN'>
INFO:gt4sd.algorithms.core:TorchDrugGCPN can not save a version based on TorchDrugSavingArguments(model_path='/Users/jab/.gt4sd/runs/', training_name='gcpn_test')
```


drugilsberg (commented):
Hi @jannisborn amazing work.

A comment on the open points.

- Regarding the property-driven training, no worries we can open an issue to track the status and merge once torchdrug merges your PR.
- Regarding `gt4sd-saving`, the only reason why the command is failing it's because the configurations for the algorithms do not overload the required method. See here in PaccMannGP for an example: https://github.com/GT4SD/gt4sd-core/blob/8addcd63c0d7622b5ef2af0d708b18b42ba142f9/src/gt4sd/algorithms/controlled_sampling/paccmann_gp/core.py#L259

jannisborn:
- `gt4sd-saving` pipeline works now, after expanding the GCPN and GraphAF algorithm configurations.
- The training pipeline can only deal simple argument types (string, float, int, bool) but not complex ones right? I removed the support for the user to pass a LR scheduler and a smiles transformation object (a callable) because there were errors in parsing these types

jannisborn:
Checks are passing, open for review @drugilsberg

jannisborn:
Closes #10

drugilsberg:
> 

That is totally correct, what I did in other pipelines is that I use the `ast` module (from standard library) for literal evaluation.
Maybe for more complex types you can do the same I did here: https://github.com/GT4SD/gt4sd-core/blob/8addcd63c0d7622b5ef2af0d708b18b42ba142f9/src/gt4sd/training_pipelines/moses/organ/core.py#L44 (not the best but functional).
Basically handling them as string and call `ast.literal_eval`.

jannisborn:
Unfortunately `ast.literal_eval` does not work with lambda, so I had to use `eval`.  See the `ast` docs:
```txt
The string or node provided may only consist of the following
    Python literal structures: strings, bytes, numbers, tuples, lists, dicts,
    sets, booleans, and None.
```

drugilsberg (approved):
Looks good to me.

PR #59, opened 2022-04-05
feat: setting up binder hosting.

drugilsberg:
Setting up Binder, waiting for the availability of a docker image.


jannisborn:
Nice, the binder setup runs smoothly. Unfortunately the kernel died when running all models together. 
I added an extra line to the dockerfile to install the notebook requirements

drugilsberg:
@jannisborn PR ready for review, everything seems to work fine, apart from the known resources limitations coming from mybinder.

jannisborn (approved):
Very nice! Last thought: we're not installing the `vcs_requirements` inside the container, so I assume that the gt4sd base image ensures that **all** of them are installed with correct versions. I selectively verified that the algorithms can be executed in the binder and for all the ones I tested it was working

drugilsberg:
> Very nice! Last thought: we're not installing the `vcs_requirements` inside the container, so I assume that the gt4sd base image ensures that **all** of them are installed with correct versions. I selectively verified that the algorithms can be executed in the binder and for all the ones I tested it was working

I believe we do, we use the conda.yml that it's ensuring they are installed :)

PR #63, opened 2022-04-15
feat: implementing moler for inference.

drugilsberg:
Implementing MoLeR inference pipeline: https://github.com/microsoft/molecule-generation

### DIFF
--- a/src/gt4sd/algorithms/generation/moler/implementation.py
+++ b/src/gt4sd/algorithms/generation/moler/implementation.py
@@ -1,0 +1,99 @@
+#
+# MIT License
+#
+# Copyright (c) 2022 GT4SD team
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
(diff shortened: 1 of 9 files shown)


src/gt4sd/algorithms/generation/moler/implementation.py
```diff
@@ -0,0 +1,99 @@
...
+        self._seed = seed
+
+    def generate(self) -> List[str]:
+        """Sample molecules using MoLeR.
+
+        Returns:
+            sampled molecule (SMILES).
+        """
+        # process scaffolds
+        valid_scaffolds = [
```
jannisborn, line 77: This only makes sure the scaffolds are molecules but not that they are scaffolds.
We could use:

```py
if (
    Chem.MolToSmiles(Chem.MolFromSmiles(scaffold), canonical=True) ==
    Chem.Scaffolds.MurckoScaffold.MurckoScaffoldSmiles(scaffold)
)
```

I think the murcko scaffold canonicalizes under the hood so we need to account for that too,

jannisborn (commented):
Interesting paper, so I like the effort to integrate it. It looks great, I just left one comment about validating the scaffolds.

The tests are failing, it seems that the weights are not yet on the COS, at least the tests seem to not find them.

Also, if I see correctly, the `list_versions` method is not yet implemented even though the tests are trying to use it.

src/gt4sd/algorithms/generation/moler/implementation.py
drugilsberg: I think they have an imprecise naming in their function, they indeed accept any mol object as scaffolds following their call stack

drugilsberg:
> 

exactly I had to jump in some calls, but I will put them on COS soon.

src/gt4sd/algorithms/generation/moler/implementation.py
jannisborn: Looking at their paper, this is a bit surprising but we should probably stick to the original implementation.

jannisborn (approved):

PR #65, opened 2022-05-03
Demo

jannisborn:
Added a demo notebook for a toy task of discovering molecules that are similar to the DDR1 inhibitor proposed by GENTRL but have an improved QED. The task might be overly synthetic, I could change to improving solubility (ESOL) instead of QED if needed?

Further minor improvements of this PR:
- [x] reinvent: method does not return invalid molecules anymore (added the `validate_item` method)
- [x] made the QED model of the RegressionTransformer available via COS and refactored the algorithm_versions to more descriptive names. 
- [x] torchdrug: registering the device to the task object (raises an error in some situations otherwise)
- [x] docstring: fixing example docstrings in Reinvent and RegressionTransformer


src/gt4sd/algorithms/conditional_generation/regression_transformer/core.py
```diff
@@ -143,7 +143,7 @@ class RegressionTransformerMolecules(AlgorithmConfiguration[Sequence, Sequence])
         An example for generating a peptide around a desired property value::
 
             config = RegressionTransformerMolecules(
-                search='sample', temperature=2, tolerance=5
+                algorithm_version='esol', search='sample', temperature=2, tolerance=5
```
drugilsberg, line 146: esol or solubility?

drugilsberg (changes requested):
I like it very much, I just see a small issue with the version naming. Make sure to uniform and reflects on the respective COS.

drugilsberg (changes requested):
I just realized we should update the regression transformer demo notebook

drugilsberg (approved):
this looks good to me, we are good to go and merge

PR #66, opened 2022-05-04
feat: ensuring sentencepiece before pytorch_lightning.

drugilsberg:
After discussing with @jannisborn, he pointed me to this  https://github.com/PyTorchLightning/pytorch-lightning/issues/11663.
Looking at the code-base I think these changes should address the segfault we were observing.

Signed-off-by: Matteo Manica <drugilsberg@gmail.com>


jannisborn (approved):
Looks good! Since this will pop up whenever we add new files that import lightning, we could consider adding a hook that adds these imports. Similar to what we do with the license headers

PR #67, opened 2022-05-04
granular trainer: search for the models in the given config by default

christofid:
Update the argument parser for the granular trainer to use the same config file to extract model's parameters if other configuration file for the models is not provided.


drugilsberg (approved):
Looks good to me, let's wait for the CI to see what happens.

PR #70, opened 2022-05-12
Improving demos

drugilsberg:


notebooks/explore-molecular-generative-models-latent-spaces.ipynb
```diff
@@ -12,7 +12,7 @@
   },
```
jannisborn, line 12: Same in this notebook, the traces are huge but the source code seems to be identical

jannisborn (approved):
Super cool, I like a lot the solubility case study for ISGC!

The changes in the three other notebooks seem to be mostly from re-running the notebooks, so we could consider a checkout before merging the PR.

notebooks/discovery-demo.ipynb
```diff
@@ -138,7 +138,7 @@
...
+       "        &lt;div class=&quot;list&quot;&gt;&lt;div class=&quot;cell&quot; data-mols2grid-id=&quot;0&quot;&gt;&lt;input type=&quot;checkbox&quot; class=&quot;position-relative float-left cached_checkbox&quot;&gt;&lt;div class=&quot;data data-Name&quot;&gt;&lt;/div&gt;&lt;a tabindex=&quot;0&quot; class=&quot;data data-img mols2grid-tooltip&quot; data-toggle=&quot;popover&quot; data-content=&quot;.&quot;&gt;&lt;/a&gt;&lt;div class=&quot;data data-QED-Val&quot;&gt;&lt;/div&gt;&lt;div class=&quot;data data-SMILES&quot; style=&quot;display: none;&quot;&gt;&lt;/div&gt;&lt;div class=&quot;data data-QED&quot; style=&quot;display: none;&quot;&gt;&lt;/div&gt;&lt;/div&gt;&lt;/div&gt;\n",
```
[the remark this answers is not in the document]
drugilsberg: I handled the logging from tensorflow differently so I would prefer to have the change saved.

notebooks/explore-molecular-generative-models-latent-spaces.ipynb
drugilsberg: Here the images were missing for some reason.

drugilsberg:
> Super cool, I like a lot the solubility case study for ISGC!
> 
> The changes in the three other notebooks seem to be mostly from re-running the notebooks, so we could consider a checkout before merging the PR.

I actually made sure to have the black styling for strings `'` to `"`, so I would keep the changes.
Now struggling with errors arising from the latest update of transformers to 4.19.0, will try to fix the builds, but in case it takes too long I will simply place an upper bound on the version.

drugilsberg:
> EDIT: Also it's not clear to me where we need `sacremoses` (which was added to the requiremets). I couldn't grep a single usage of it

This was done to fix a test bug introduced by transformers 4.19.0, that uses it in imports and does not install it anymore, see the commit message and the previously failed build for more info.

PR #76, opened 2022-05-24
Ci setup

### LINKED ISSUE #72: package_data not included in source distribution.
Currently on pypi we only publish a source distribution, and it's not containing the package_data. It's there when building a wheel, e.g. with `python setup.py bdist_wheel`.

While building a wheel is something we should anyway do, we also need to add a [MANIFEST.in](https://packaging.python.org/en/latest/guides/using-manifest-in/) file.
Edit: it should be possible without MANIFEST.in

C-nit:
https://setuptools.pypa.io/en/latest/userguide/datafiles.html:
> If using the include_package_data argument, files specified by package_data will not be automatically added to the manifest unless they are listed in the [MANIFEST.in](https://packaging.python.org/en/latest/guides/using-manifest-in/) file or by a plugin like [setuptools-scm](https://pypi.org/project/setuptools-scm) or [setuptools-svn](https://pypi.org/project/setuptools-svn).

https://github.com/GT4SD/gt4sd-core/runs/6570553643?check_suite_focus=true#step:6:103 confirms e.g. the py.typed file finally being available in the distributions.

Closes #72.

@drugilsberg This branch released a .dev prerelease version. After merge I will push a bump to 0.36.2 with a tag, and will pull 0.36.1 from pypi and delete the tag.

### DIFF
--- a/.github/workflows/pypi.yaml
+++ b/.github/workflows/pypi.yaml
@@ -28,25 +28,13 @@ jobs:
         uses: actions/setup-python@v1
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install wheel
+      - name: Install build
         run: |
-          python -m pip install wheel
+          python -m pip install build
           python -m pip freeze
       - name: Build dists
         run: |
-          python setup.py sdist bdist_wheel
-      - name: Update build dependencies
-        run: |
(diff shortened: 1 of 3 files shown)


drugilsberg (approved):
LGTM

PR #77, opened 2022-05-25
Multiproperty support in Regression Transformers

jannisborn:
Supporting multiproperty prediction and multiproperty conditional generation in the Regression Transformer

- [x] added a new model `logp_and_synthesizability` that has been trained on pubchem for logp and SCScore
- [x] ensuring proper handling of property scales: The model has been trained on logp in a [0,1] scale but on the raw scale for SCScore (1,5). The code is flexible in that the (optional) scaling parameters are read from the model config, s.t. the user only operates on the intuitive, raw scale without ever seeing the scaled values.
- [x] return types of property prediction change from float to string, to reflect multiproperty nature. Instead of `0.123` it will be `"<qed>0.123"`. I chose strings as return types because dicts are not hashable which is a blocker in the sampling step (where we remove duplicates)
- [x] adding wrappers for random masking of tokens and for only masking specific atom types (with certain probability)
- [x] added pytest for `logp_and_synthesizability` model and also for the masking wrapper
- [x] improved error handling and docstrings


jannisborn:
Completed all duties above. We can now generate samples like this:

```py
target = 'CCOC1=NC=NC(=C1C)NCCOC(C)C'
    algorithm_version='logp_and_syfsdnthesizability', search='sample', tolerance=15,
    sampling_wrapper={
        'property_goal': {'<logp>': 6.65, '<scs>': 3.6}, 'fraction_to_mask': .4, 'tokens_to_mask': ['C']
    }
)
)
```
This would start from the target SMILES as seed, randomly mask 40% of the carbon atoms (`[C]` and `[=C]` in the SELFIES) and prime the model with property constraints based on the desired logP and synthesizability. As before, `tolerance` is given in percentage of deviation in the predicted properties of the generated molecules.

drugilsberg (approved):
Did not have time for an in-depth review but it looks good, let's wait for the CI to finish.

PR #78, opened 2022-05-27
Reinvent seq length

jannisborn:
Added support for restricting the sequence lengths of the REINVENT genreators.


drugilsberg (approved):
LGTM

PR #81, opened 2022-06-10
cli-upload

georgosgeorgos:
#### cli-upload

Add `upload` functionality to the command line. It gives the user the possibility to upload specific artifacts on a server.

Given a specific `version` for an algorithm:

* check if that version is already on the server:
                 - check if the folder `bucket/algorithm_type/algorithm_name/algorithm_application/version/` exists.
* If yes, tell the user and stop the upload.
* If not, upload all the files in that version.

cli-upload relies on `minio` and has been tested locally using `docker-compose`. 
cli-upload can be used to upload on a cloud or local server.

-----------

#### How to use cli-upload

Following the example in the README (in the `Saving a trained algorithm for inference via the CLI command` section) and assuming a trained model in `/tmp/test_cli_upload`, run:

`gt4sd-upload --training_pipeline_name paccmann-vae-trainer --model_path /tmp/test_cli_upload --training_name fast-example --target_version fast-example-v0 --algorithm_application PaccMannGPGenerator`


requirements.txt
```diff
@@ -22,3 +22,5 @@ tensorflow==2.1.0
 keras==2.3.1
 sacremoses>=0.0.41
 protobuf<=3.20.1
+multiprocess==0.70.12.2
```
drugilsberg, line 25: Why are we pinning this? From the PR it is not immediately obvious. If we really need this, I would add it in setup.cfg and I would relax the requirements if possible.

src/gt4sd/algorithms/core.py
```diff
@@ -442,6 +443,34 @@ def list_versions(cls) -> Set[str]:
         versions = versions.union(get_algorithm_subdirectories_in_cache(prefix))
         return versions
 
+    # QUESTION: should this be in core.py? Uploading to S3 is not a core functionality.
```
drugilsberg, line 446: in principle I agree with the comment, but listing remote versions I think it's useful. Maybe we should rename it thinking about the future and the possibility of using other types of remote storage systems. My main question here is, how this differ from `list_versions` besides the function signature?
If they do not introduce any change in the logic I would call one inside the other to reduce code duplication.

src/gt4sd/s3.py
```diff
@@ -147,6 +159,46 @@ def sync_folder(
...
+        client = GT4SDS3Client(
+            host=host, access_key=access_key, secret_key=secret_key, secure=secure
+        )
+        logger.info("starting syncing")
+        client.upload_file(bucket, target_filepath, source_filepath)
+        logger.info("syncing complete")
+    except Exception:
+        logger.exception("generic syncing error")
+        raise S3SyncError(
+            "CacheSyncingError",
```
drugilsberg, line 197: I would change the error type/name to 'UploadArtifactsErrors'

src/gt4sd/s3.py
```diff
@@ -147,6 +159,46 @@ def sync_folder(
...
+            f"error in uploading path={target_filepath} with host={host} access_key={access_key} secret_key={secret_key} secure={secure} bucket={bucket}",
```
drugilsberg, line 198: In principle I would avoid logging secrets, for the open ones that is totally ok, but in case of custom deployments of storage it may exposed sensitive secrets.

drugilsberg (changes requested):
@georgosgeorgos amazing work, see my comments, mostly minor changes/refactor and points of discussion.

src/gt4sd/algorithms/core.py
georgosgeorgos: In `list_versions` we search on s3 (`get_algorithm_subdirectories_with_s3`) and locally (`get_algorithm_subdirectories_in_cache`) for versions. 
In `list_versions_s3` we search only s3 because we could have already saved our version in the cache (for example using cli-saving before cli-upload) but still want it on s3.

We can call one inside the other but then we need two outputs: one set with the versions on s3 and another set with all the versions (s3 and cached).

drugilsberg: ah good catch sorry I missed it, then I would simply rename `list_versions_s3` to `list_remote_versions`.

georgosgeorgos:
For the moment we don't have a direct way (for example `os.environ`) to use environment variables in [configuration.py](https://github.com/GT4SD/gt4sd-core/blob/main/src/gt4sd/configuration.py#L39)

drugilsberg:
> For the moment we don't have a direct way (for example `os.environ`) to use environment variables in [configuration.py](https://github.com/GT4SD/gt4sd-core/blob/main/src/gt4sd/configuration.py#L39)

The BaseSettings from pedantic are directly initialized from the env variables (where the upper case env var is read into the corresponding lower case class attribute).
To access those simply add the the to the configuration object and instantiate it. I hope this helps.

src/gt4sd/algorithms/core.py
```diff
@@ -442,6 +443,61 @@ def list_versions(cls) -> Set[str]:
...
+    @classmethod
```
drugilsberg, line 447: we should keep just one of the two methods

docs/source/gt4sd_server_upload_md.md
```diff
@@ -0,0 +1,115 @@
...
+
+After running `docker compose up` inside localhost-server this script will create a local minio server and the bucket `myminio` structure .
+If everything is working you should be able to see `a_file.txt` running:
+
+```
+mc ls local-dev/gt4sd-algorithms-artifacts/
+```
+-------
+
+## Upload models
```
drugilsberg, line 113: here we should add a mention that the env variables need to be configured approriately

requirements.txt
```diff
@@ -22,3 +22,5 @@ tensorflow==2.1.0
 keras==2.3.1
 sacremoses>=0.0.41
 protobuf<=3.20.1
+#multiprocess==0.70.12.2
```
drugilsberg, line 25: ig not needed I would remove it all the way from here

drugilsberg (changes requested):
We are getting there :)

docs/source/gt4sd_server_upload_md.md
georgosgeorgos: Added a line to clarify that the user can upload only after setting env variables and following steps 1-4.

drugilsberg:
> > The BaseSettings from pedantic are directly initialized from the env variables (where the upper case env var is read into the corresponding lower case class attribute). To access those simply add the the to the configuration object and instantiate it. I hope this helps.
> 
> Fixed it. Thanks.
> 
> When using upload functionality locally, the server can be `http` and we need `gt4sd_s3_secure=False` in [configuration.py](https://github.com/GT4SD/gt4sd-core/blob/main/src/gt4sd/configuration.py#L53). I tried setting an environment variable `GT4SD_S3_SECURE="False"` and then use `gt4sd_s3_secure: bool = distutils.util.strtobool("True")` as default in [configuration.py](https://github.com/GT4SD/gt4sd-core/blob/main/src/gt4sd/configuration.py#L53). This works, but raises an error in one of the checks and I removed it. So at the moment the user has to manually change the value in configuration.py

in principle the strtobool is not needed, you can simply set `gt4sd_s3_secure: bool = True` and define `export GT4SD_S3_SECURE-false` and it should work right away.

requirements.txt
georgosgeorgos: multiprocess updated to `0.70.13` two weeks ago. 
With the new version I had problems using `gt4sd`. 
Downgrading to `0.70.12.2` everything works fine.

README.md
```diff
@@ -268,6 +268,17 @@ Run the algorithm via `gt4sd-inference` (again the model produced in the example
...
 
+### Uploading a trained algorithm on a server via the CLI command
+
+If you have access to a server (local or cloud) you can upload your trained models easily. The syntax follow the saving pipeline using `gt4sd-upload`:
+
+```sh
+gt4sd-upload --training_pipeline_name paccmann-vae-trainer --model_path /tmp/gt4sd-paccmann-gp --training_name fast-example --target_version fast-example-v0 --algorithm_application PaccMannGPGenerator
+```
+
+You have to update `gt4sd_s3_host`, `gt4sd_s3_access_key`, `gt4sd_s3_secret_key`, `gt4sd_s3_secure` and `gt4sd_s3_bucket` appropriately to upload the models.
```
drugilsberg, line 279: I would refer to the upper case env variables since they will overrwrite the configuration values.

georgosgeorgos:
There is a problem with GitHub Action related to this issue - https://github.com/actions/runner/issues/667. Thanks @jannisborn for finding it.

drugilsberg (approved):
Added a minor rephrasing in the docs. Good to go, besides the CLA issue.

Issue #80, opened 2022-06-09
gt4sd-trainer and pytoda error


georgosgeorgos:
### Contributor License Agreement

- [X] I accept and sign the contributor license agreement ([CLA](../../CONTRIBUTING.md#contributor-license-agreement)) to contribute to GT4SD.

jannisborn:
This problem is caused by side-effects from torchdrug: https://github.com/DeepGraphLearning/torchdrug/issues/96

It's unclear how it can be fixed, so we have to stall this until we hear from torchdrug developers

PR #84, opened 2022-06-13
Patch saving example

jannisborn:
Switch the CLI example from `PaccMannGP` to Moses `VAEGenerator` to go around #80.
It's not a stable fix though, the issue related to PaccMannGP persists


drugilsberg:
Thanks for the hot fix, I'm afraid we might need to patch the pipelines using pytoda to overcome the torchdrug redefinition of torch modules. Maybe something along these lines: https://github.com/GT4SD/gt4sd-core/blob/2f049080a6f4132fad6d7834b57c64a38e96be7e/src/gt4sd/algorithms/generation/torchdrug/implementation.py#L52.

drugilsberg (approved):
Good job, see the comment.

jannisborn:
Thanks! Yes I was thinking in a similar direction but since all affected torch classes are **overwritten** and the original classes are not saved anywhere (unlike in the `torch.nn` case), I dont see any way to retrieve them, especially also since `importlib.reload` does not restore functionality.

See my comment here: https://github.com/DeepGraphLearning/torchdrug/issues/96#issuecomment-1150814220

drugilsberg:
@jannisborn totally get this, the only solution I see to preserve subclassing is to ensure the imports of all the relevant classes before `torchdrug` gets imported.
See below:

```console
$  ipython
Python 3.7.13 (default, Mar 28 2022, 07:24:34) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.32.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from torch.utils.data.dataset import ConcatDataset, Dataset

In [2]: issubclass(ConcatDataset, Dataset)
Out[2]: True

In [3]: import torchdrug

In [4]: issubclass(ConcatDataset, Dataset)
Out[4]: True

In [5]:   # Ctrl + d                                                                                                                                                                                                                                    
Do you really want to exit ([y]/n)? y
$  ipython
Python 3.7.13 (default, Mar 28 2022, 07:24:34) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.32.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import torchdrug

In [2]: from torch.utils.data.dataset import ConcatDataset, Dataset

In [3]: issubclass(ConcatDataset, Dataset)
Out[3]: False

In [4]: 
```

jannisborn:
great that seems very promising! I will check this later in detail. Let's roll back to a draft PR here then

PR #86, opened 2022-06-14
Torchdrug dataset fix

jannisborn:
Fixing #80 by ensuring the import of the sane torch datasets. After torchdrug is imported and breaks the dataset classes, we re-assign the sane datasets to them.

The same procedure could be done for other patches in `torchdrug/patch.py`, e.g., torch Optimizers and LRSchedulers.

(I also refactored the `num_relation` argument to align across both TorchDrug model, one model mistakenly had a trailing s)


drugilsberg (approved):
Great job Jannis

PR #94, opened 2022-06-20
feat: add public hub with writing access.

georgosgeorgos:
Add public hub.

* Expose an additional access and secret key pair with writing access on the `gt4sd-cos-hub-algorithms-artifacts` bucket.
* `sync` and `list` versions using all the remote buckets (public read-only + public hub).
* `upload` only on the public hub.

Signed-off-by: Giorgio Giannone <giorgio.giannone@ibm.com>


src/gt4sd/configuration.py
```diff
@@ -141,6 +160,7 @@ def get_cached_algorithm_path(prefix: Optional[str] = None) -> str:
 def get_algorithm_subdirectories_with_s3(prefix: Optional[str] = None) -> Set[str]:
```
drugilsberg, line 160: just noticed we are missing a docstring, let's add it in this PR.

src/gt4sd/configuration.py
```diff
@@ -149,7 +169,27 @@ def get_algorithm_subdirectories_with_s3(prefix: Optional[str] = None) -> Set[st
             host=host, access_key=access_key, secret_key=secret_key, secure=secure
         )
         bucket = gt4sd_configuration_instance.gt4sd_s3_bucket
-        return client.list_directories(bucket=bucket, prefix=prefix)
+
```
drugilsberg, line 172: what about creating a method returning the directories taking as input the parameters needed to instantiate and query the client?

```python
def get_algorithm_subdirectories_from_s3_coordinates(prefix: Optional[str] = None, host: str, access_key: str, secret_key: str, bucket_name: str, secure: bool = True) -> Set[str]:
    clibent = GT4SDS3Client(
            host=host,
            access_key=access_key,
            secret_key=secret_key,
            secure=secure,
        )
        return client.list_directories(bucket=bucket_name, prefix=prefix)
```

And then:

```python
...
dirs = get_algorithm_subdirectories_from_s3_coordinates(...)
dirs_hub = get_algorithm_subdirectories_from_s3_coordinates(...)
versions = dirs.union(dirs_hub)
return versions
```

drugilsberg (commented):
Minor comment on a potential refactor

georgosgeorgos:
I left the original env variables (without `_HUB`) in the README where we explain how to set a local server.

drugilsberg:
> variables (without `_HUB`) in the README where we explain how to set a local

why not updating the README.md to simply show how to upload to the public hub, and delegate the local setup to the relevant doc page? I think it would be more effective for users going through the docs.

PR #95, opened 2022-06-21
Code example using PaccMannGP

jannisborn:
I set up an `examples` folder that can be expanded with other applications but for the moment it only contains an example of `PaccMannGP` with different target functions.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -6,5 +6,5 @@ authors:
 title: "GT4SD (Generative Toolkit for Scientific Discovery)"
 version: 0.40.0
 url: "https://github.com/GT4SD/gt4sd-core"
-# doi: TBD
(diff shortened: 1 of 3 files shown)


CITATION.cff
```diff
@@ -4,7 +4,7 @@ authors:
   - family-names: Team
     given-names: GT4SD
 title: "GT4SD (Generative Toolkit for Scientific Discovery)"
-version: 0.22.0
+version: 0.40.0
 url: "https://github.com/GT4SD/gt4sd-core"
 # doi: TBD
```
drugilsberg, line 9:
```suggestion
doi: "https://doi.org/10.5281/zenodo.6673578"
```

drugilsberg (commented):
Good idea, we can even add python scripts to it. I created a zenodo DOI.

drugilsberg (approved):

PR #99, opened 2022-06-23
Constant token length in RT notebooks

jannisborn:
Fixing error in notebooks caused by #77


drugilsberg (commented):
Thanks for the fix. Do we need to apply also to [this notebook](https://github.com/GT4SD/gt4sd-core/blob/main/notebooks/discovery-demo.ipynb)?

jannisborn:
Fortunately no, because in there the QED is already represented with 5 tokens

PR #101, opened 2022-06-27
Training pipeline Regression Transformer

jannisborn:
Adding new training pipeline

- allows to finetune existing models available in the toolkit
- allows to train models from scratch


drugilsberg:
> PR is complete. One aspect that might require attention in the future is our argument parsing from the CLI.
> 
> As far as I can see, the integrated pipeline is the first one that exposes the full power of `transformers.training_args.TrainingArguments` to the user. All CLI arguments that are listed can be specified by the user and are passed down to the HF Trainer. For a full list see here: https://github.com/microsoft/huggingface-transformers/blob/master/src/transformers/training_args.py#L65
> 
> This is a nice new feature that we should explore for other pipelines as well, but we have to keep a few things in mind:
> 
>     1. Our CLI parser cannot handle non-trivial types like Iterables. If an argument is annotated as an `Iterable` the CLI parser raises an error. This is in contrast to the HF Trainer, see e.g. [`label_names` argument here](https://github.com/microsoft/huggingface-transformers/blob/b349167cdd8428dff5600a8635dd1138045fa623/src/transformers/training_args.py#L490). Current solution: Overwrite these arguments and require them to be strings which effectively makes them unusable. This only concerns arguments that are irrelevant for the RT but we would need a logic in the future to convert a string like `label1,label2` into `["label1", "label2"]`.
> 
>     2. Arguments that are specified as empty strings are converted to `None` by the `gt4sd.cli.ArgumentParser`. This is the normal behavior of the `argparse` package but what is problematic is how our CLI inputs are parsed into the argument classes. Some HF arguments have empty strings as default and the `HFArgumentParser` requires them to be strings and not `None`. Solution: Reassigning to empty strings in `__post_init__`.
> 
>     3. One HF argument does not have a default, namely `output_dir`. This is a bit annoying to handle in the unittests where we create the parser first with all defaults and then overwrite the defaults with the values from the dictionary hardcoded in the test file.

Great job, I will check this carefully, but I have a comment on point 1. For the time being in other pipelines I mitigated this effect by accepting a string containing ","-separated values or ast values and parsing this after initialization. Not the best solution, but it should enable to train models using iterable-like args.

jannisborn:
The tests pass locally with the same transformers version now. The current CI error is caused by argparse not liking one of the HF default training arguments. I cant change the argparse version/reproduce the error locally 🤔

src/gt4sd/cli/argument_parser.py
```diff
@@ -181,7 +181,7 @@ def _add_dataclass_arguments(self, dtype: DataClassType) -> None:
                 kwargs["type"] = partial(none_checker, dtype=field.type.__args__[0])
                 assert all(
                     x == kwargs["type"] for x in field.type.__args__
-                ), f"{field.name} cannot be a List of mixed types"
+                ), f"{field.name} cannot be a List of mixed types: {field.type.__args__} with"
```
drugilsberg, line 184:
```suggestion
                ), f"{field.name} cannot be a List of mixed types: {field.type.__args__}"
```

src/gt4sd/cli/load_arguments_from_dataclass.py
```diff
@@ -103,7 +109,7 @@ def extract_fields_from_class(
                 raise ValueError(f"{arg_fields[field_name]['type']} not supported")
 
         else:
-            raise ValueError(
+            logger.error(
```
drugilsberg, line 112: why not raising as done above?

src/gt4sd/training_pipelines/regression_transformer/core.py
```diff
@@ -0,0 +1,152 @@
...
+#
+"""Regression Transformer training utilities."""
+import os
+from dataclasses import dataclass, field
+from typing import Optional
+
+from ..core import TrainingPipelineArguments
+from .utils import TransformersTrainingArgumentsCLI
+
+DATA_ROOT = os.path.join(
```
drugilsberg, line 32: cache is dynamic see here: https://github.com/GT4SD/gt4sd-core/blob/09a7b1d34233f737fa763d0b23a3fab1cc9b4e29/src/gt4sd/configuration.py#L46.
We should get the path from the settings.

drugilsberg (changes requested):
I think the work is great, but there is something that we should change.
Right now we prepare train and data splits in the trainer and this is not very practical.
Also compared to other pipelines it's implementing a different logic.
I would refactor the pipeline to get two dataset filepaths one for training and one for testing/val.
In this way we can;
- get rid of the splitting logic in the pipeline
- have a uniform behaviour compared to other trainers.

src/gt4sd/cli/argument_parser.py
jannisborn: Well spotted!

src/gt4sd/cli/load_arguments_from_dataclass.py
jannisborn: If we raise, the tests in `test_training_pipelines.py` fail with errors like:
```sh
 ValueError:  argument evaluation_strategy: <enum 'IntervalStrategy'> not supported
```
Our parsers are quite restrictive on supported argument types for the dataclasses which conflicts with the HF Training arguments that support arbitrary types: https://github.com/huggingface/transformers/blob/6aae59d0b54f04c13a79f80b708622db8e8a17e4/src/transformers/training_args.py#L496


IMHO this is solved incorrectly in HF because the type annotation should be `Union[str, IntervalStrategy]` rather than `IntervalStrategy`. But since the RT trainer embraces all HF training arguments this type annotation is propagated down and raises in the tests. I think the best way would be to solve this upstream in HF. I can prepare a PR there, WDYT?

jannisborn:
About the data preparation:
I think it's important that the pipeline can handle csv files with text and arbitrarily many properties. A common usecase is SELFIES and we should not expect that users prepare data in the correct format for the RT.

But I agree to your point about the splitting.
I can refactor so that the user provides `train_data_path` and `test_data_path` which are both `.csv` files with text and numerical properties. So the splitting is outsourced to the user but the csv -> RT format conversion is still done by the TrainerPipeline.

src/gt4sd/cli/load_arguments_from_dataclass.py
drugilsberg: I think this would be an ideal solution, may I suggest leaving a comment before the `logger.error` saying something like:
```
# NOTE: not raising since the HF training args might introduce typing inconsistencies
```

drugilsberg:
> perties. So the splitting is outsourced to the user but the csv -> RT format conversion is still done by the TrainerPipeline.

This sounds like a perfect solution, I think it should do the trick.

src/gt4sd/cli/load_arguments_from_dataclass.py
jannisborn: PR opened here: https://github.com/huggingface/transformers/pull/17934

But I'm not sure this can be merged given the HF and argparse requirements, but I will try.

jannisborn: Added the comment, as you suggested see the last commit

requirements.txt
```diff
@@ -1,7 +1,7 @@
 # pypi requirements
 minio==7.0.1
 numpy>=1.16.5
-torch>=1.0
+torch>=1.0,<=1.11.0
```
drugilsberg, line 4: @jannisborn is it something we strongly need? I would avoid upper bound as much as possible as they open the gates for deps hell.

jannisborn: I know and I agree. But `torch==1.12` introduced a refactor of the dataset classes that moved the `*DataPipe` stuff here: https://github.com/GT4SD/gt4sd-core/blob/main/src/gt4sd/training_pipelines/torchdrug/unpatch.py#L33
to another submodule. 

It's the only way I see to guarantee a stable release. Alternatively, we could remove the 3 affected classes from the `unpatch.py` and would accept that user's downstream code might break (i dont think we use these classes ourselves anywhere so our own CI should be unaffected)

jannisborn: Or we have an if/else in the import based on the torch version

drugilsberg: I know it's not the best, but since it will be mostly confined to the unpatch method let's put the if/else.
We can use: `pkg_resources.get_distribution("torch").version` and then compare using:

```python
import pkg_resources
from packaging import version
if version.parse("pkg_resources.get_distribution("torch").version) > version.parse(""1.11"):
    pass
else:
   pass
```

drugilsberg (approved):
Looks good, let's see what happens with the checks. In case they pass, we can merge and release.

PR #103, opened 2022-06-28
Add customizable tensoboard logger

christofid:
This PR:

- add customizable tensoboard logger to control the output directory.
- add GranularPytorchLightningTrainingArguments class to use the same default values for the training parameters as the ones defined in Granular's framework.


drugilsberg (approved):
Looks good, @christofid do you think #101 should adress the mypy problem appearing in the checks?

PR #105, opened 2022-07-01
Property submodule

jannisborn:
First sketching of a submodule `gt4sd.properties` that can contain domain-specific submodules like `gt4sd.properties.molecules` and `gt4sd.properties.proteins`.

This is a very crude draft to initiate a discussion where this should go.

At the moment, the `protein` and the `molecule` submodule implement a lot of property functions that can be used to evaluate properties. I'm aiming for an unified interface where we have a single factory like `MOLECULE_FACTORY` that can be queried with property keys (e.g., `"penalized_logp"`) and return a `Callable[[SmallMolecule], Property]`.

Done:
- [x] Implement the `protein` submodule with access to all `modlamp` protein properties (hydrophobicity, stability etc). Can be called with AAS or rdkit mol objects
- [x] Implement the `molecule` submodule with access to property functions from `guacamol` and `paccmann_generator` and some functions of TDC. Can be called with SMILES or rdkit mol objects

To discuss:
- [x] How does this submodule relate to `gt4sd.domains.materials`? It might be nice to unify the interfaces since every property could in principle be usable as a scorer
- [x] We could make the `gt4sd.algorithms.property_predictors`, like the AffinityPredictor, available here too
- [x] The complex property functions like predicting toxicity will have to be integrated into the ApplicationRegistry like the generative models. 
- [ ] TDC: I only added support for measuring similarity to a seed molecule. They have a lot of very specific oracles (https://github.com/mims-harvard/TDC/blob/main/tdc/oracles.py#L110) where I'm not convinced they are useful so I spared them out. We should consider integrating their models to measure affinity for specific targets (I included an example with `activity_against_target`) but I'm not sure we want that because 1) it will require pinning to old sklearn version (<1) and 2) we could use our generic affinity predictor instead (even though this requires that people have the protein sequence at hand)
- [x] TDC: We should go through the existing oracle carefully. Many of those require an API key but assuming that they work out of the box, I think it would be great to make those available via `gt4sd` too.


drugilsberg:
> * ity predictor instead (even though this requires that people have the protein sequence at hand)
> 
> *

Love the initiative and I agree on most of the points. The factory idea is the model where we should go, I would consider even extending the concept to support parameters:

```python
MOLECULE_FACTORY: Dict[str, Tuple[Callable[[Molecule, Parameters], Property], ParametersSchema]]
```

In this way we can have arbitrarily complex calls.

I would move in the direction of having complex predictors relying on a separate cache and either "promote" some of them under `gt4sd.algorithms.property_predictors`, or get ride of that submodule all the way and focusing on `gt4sd.properties`.

Same reasoning for `gt4sd.domains.materials` I would unify as much as possible.

requirements.txt
```diff
@@ -22,5 +22,5 @@ tensorflow==2.1.0
 keras==2.3.1
 sacremoses>=0.0.41
 protobuf<=3.20.1
-
-
+modlamp>=4.0.0
+PyTDC>=0.3.6
```
drugilsberg, line 26:
```suggestion
PyTDC>=0.3.6

```

src/gt4sd/properties/molecules/core.py
```diff
@@ -0,0 +1,238 @@
+from typing import Callable
```
drugilsberg, line 1: All docstrings need to be completed with `Args` and `Returns`

src/gt4sd/properties/molecules/core.py
```diff
@@ -0,0 +1,238 @@
...
+    """
+    return similarity_fn(to_smiles(mol))
+
+
+def activity_against_target(
+    mol: SmallMolecule, affinity_fn: Callable[[SmallMolecule], Property]
+) -> Property:
+    """Calculate the activity of a molecule against a target molecule.
+
+    For example:
```
drugilsberg, line 222: For examples we have a specific style to follow for rendering in sphinx, see PaccMannRL generators for examples.

src/gt4sd/properties/molecules/core.py
```diff
@@ -0,0 +1,238 @@
+from typing import Callable
+
```
drugilsberg, line 2: In general I would turn all of those in classes with `__call__` method and having parameters in the `__init__` following a common class interface. In this way all functions can be called in the same way.

src/gt4sd/properties/molecules/core.py
```diff
@@ -0,0 +1,238 @@
...
+    """The amount of heterocycles of a molecule."""
+    return CalcNumHeterocycles(to_mol(mol))
+
+
+def number_of_stereocenters(mol: SmallMolecule) -> int:
+    """The amount of stereo centers of a molecule."""
+    return CalcNumAtomStereoCenters(to_mol(mol))
+
+
+def similarity_to_seed(
```
drugilsberg, line 200: see comment  on classes

src/gt4sd/properties/molecules/core.py
```diff
@@ -0,0 +1,238 @@
...
+
+    func = get_activity_fn(target_mol='DRD2')
+    activity_against_target(mol='CCC', affinity_fn=func)
+    ```
+
+    """
+    return affinity_fn(to_smiles(mol))
+
+
+# TODO: Need to put the trained models on COS and implement caching logic
```
drugilsberg, line 234: This is something that should be relatively easy if we introduce a dedicated sub-bucket path.

src/gt4sd/properties/proteins/core.py
```diff
@@ -0,0 +1,99 @@
+from ...domains.materials import MacroMolecule
```
drugilsberg, line 1: same comments as for small molecules files.

src/gt4sd/properties/molecules/core.py
drugilsberg, line 1: I would name this one`small_molecules.py` or `rdkit.py`

drugilsberg (commented):
I left some comments, the work look already really nice.

src/gt4sd/properties/molecules/core.py
georgosgeorgos: What do you mean here? Should we use: 

```
T = TypeVar("T", Protein, Omics)
S = TypeVar("S", bound=SmallMolecule)
```

georgosgeorgos:
Now we have a common interface and we can compute all the properties using the `callable_fn` except for `similarity_seed` and `activity_against_target`.

jannisborn:
- Condensed parameter handling (for those callables that dont have hyperparameters)
- Functional implementations of `similarity_seed` and `activity_against_target`
- renamed `PropertyPredictor` to `Property` since many properties are **not** predicted
- distinguished between `CallableConfigurableProperty` and `CallableProperty` where the latter is static after constructions and the former can be called at execution time with hyperparameters
- lowered scikit-learn upper limit to reflect TDC oracle requirements

TODO:
- [ ] Unit tests
- [ ] Loading from JSON: we need to implement `from_json` to restore a `CallableProperty` from a json. I'm not exactly sure how to ensure a safe mapping to the keys in the FACTORY (we use different snake case for the keys but the class names are  in camel case). Any idea @drugilsberg, maybe like you did it in the registry?

drugilsberg (approved):
LGTM

drugilsberg:
> * `gt4sd.algorithms.property_predictors`

ticking the ones for which I opened issues

PR #111, opened 2022-07-12
fix: flexible device usage in RT inference

jannisborn:
Ensuring that no device mismatches occur if RT inference is executed with CUDA availability


drugilsberg (commented):
Minor refactor suggestion.

PR #114, opened 2022-07-13
Patches

jannisborn:
Some minor patches for RT device handling and torchdrug unpatching


drugilsberg (approved):
Great job, I also like better the patch refactor.

Issue #41, opened 2022-02-22
Add dedicated interface for predictive models


drugilsberg:
Currently predictive models such as PaccMann, adhere to the GenerativeAlgorithm interface. This is not an ideal fit, we should consider adding a dedicated interface.

jannisborn:
Started working on this in #105

jannisborn:
Completed via #121

PR #121, opened 2022-08-11
Artifact storage for property predictors

### LINKED ISSUE #116: Handle artifacts for property predictors
Currently we excluded some property predictors because they have complex logics or they rely on artifacts (e.g., toxicity)

We should make those available, most likely relying on S3 syncing, we can have a separate bucket for those.

jannisborn:
Closes #116

Now we can store artifacts also for property predictors

- New property predictors are tested 
- One thing that remains to do is to have functions under `gt4sd.properties.molecules.functions`. Atm this is not yet supported since it would yield circular imports.

### DIFF
--- a/src/gt4sd/algorithms/core.py
+++ b/src/gt4sd/algorithms/core.py
@@ -747,7 +747,7 @@ class AlgorithmConfiguration(Generic[S, T]):
             )
 
     @classmethod
-    def ensure_artifacts_for_version(cls, algorithm_version: str) -> str:  # type: ignore
+    def ensure_artifacts_for_version(cls, algorithm_version: str) -> str:
         """The artifacts matching the path defined by class attributes and the given version are downloaded.
 
         That is all objects under ``algorithm_type/algorithm_name/algorithm_application/algorithm_version``
@@ -921,6 +921,7 @@ class PropertyPredictor(ABC, Generic[S, U]):
             context: the context in which a property of an item can be
                 computed or checked is very application specific.
         """
+        logger.warning("This class will be deprecated in a future release.")
         self.context = context
         # or pass these with methods?
 
--- a/src/gt4sd/algorithms/core.py
+++ b/src/gt4sd/algorithms/core.py
@@ -73,6 +73,8 @@ U = TypeVar("U")  # used for additional context (e.g. part of target definition)
 Targeted = Callable[[T], Iterable[Any]]
 # callable not taking any target
 Untargeted = Callable[[], Iterable[Any]]
+# predictive model
+Predictor = Callable[[Any], Any]
 
 
 class GeneratorAlgorithm(ABC, Generic[S, T]):
@@ -277,6 +279,94 @@ class GeneratorAlgorithm(ABC, Generic[S, T]):
         """
         logger.info("no parameters validation")
         return configuration
+
+
+class PredictorAlgorithm(ABC, Generic[S, T]):
+    """Interface for automated prediction via an :class:`AlgorithmConfiguration`."""
+
+    #: The maximum amount of time we should let the algorithm run
+    max_runtime: int = gt4sd_configuration_instance.gt4sd_max_runtime
+
+    def __init__(self, configuration: AlgorithmConfiguration[S, T]):
+        """Targeted or untargeted generation.
+
+        Args:
+            configuration: application specific helper that allows to setup the
+                generator.
+        """
+        logger.info(
+            f"runnning {self.__class__.__name__} with configuration={configuration}"
+        )
+        self.configuration = configuration
+        self.predictor = self.get_predictor(configuration)
+
+    def get_predictor(
+        self,
+        configuration: AlgorithmConfiguration[S, T],
+    ) -> Predictor:
+        """Set up the predictive model from the configuration. This is called at
+            instantiation of the PredictorAlgorithm and must return a callable.
+
+        Returns:
+            predictor, a callable that takes an item and returns a prediction.
+        """
+        logger.info("ensure artifacts for the application are present.")
+        self.local_artifacts = configuration.ensure_artifacts()
+        model: Predictor = self.get_model(self.local_artifacts)
+        return model
+
+    @abstractmethod
+    def get_model(self, resources_path: str) -> Predictor:
+        """
+        Restore the model from a local path.
+
+        Note:
+            This is the major method to implement in child classes, it is called
+            at instantiation of the PredictorAlgorithm and must return a callable:
+
+        Args:
+            resources_path: local path to the downloaded artifacts.
+
+        Returns:
+            Predictor (callable)
+        """
+        raise NotImplementedError("Not implemented in baseclass.")
+
+    def timeout(self, signum, frame):
+        raise TimeoutError(
+            "Alarm signal received, probably because a signal.alarm timed out.",
+        )
+
+    def predict(self, input: Any) -> Any:
+        """Perform a prediction for an input.
+        Args:
+            input: the input for the predictive model
+
+        Raises:
+            TimeoutError: when the walltime limit is hit.
+
+        Returns:
+            the prediction.
+        """
+
+        signal.signal(signal.SIGALRM, self.timeout)
+        signal.alarm(self.max_runtime)
+
+        try:
+            predicted = self.predictor(input)
+        except TimeoutError:
+            detail = (
+                f"Predicting took longer than maximum ({self.max_runtime} seconds)."
+            )
+            logger.warning(detail + " Exiting now!")
+        except Exception:
+            raise Exception(f"{self.__class__.__name__} failed with {input}")
+        signal.alarm(0)
+        return predicted
+
+    def __call__(self, input: Any) -> Any:
+        """Alias for `self.predict`."""
+        return self.predict(input)
 
 
 @dataclass
@@ -657,7 +747,7 @@ class AlgorithmConfiguration(Generic[S, T]):
             )
 
     @classmethod
-    def ensure_artifacts_for_version(cls, algorithm_version: str) -> str:
+    def ensure_artifacts_for_version(cls, algorithm_version: str) -> str:  # type: ignore
         """The artifacts matching the path defined by class attributes and the given version are downloaded.
 
         That is all objects under ``algorithm_type/algorithm_name/algorithm_application/algorithm_version``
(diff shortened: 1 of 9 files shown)


jannisborn:
Failing tests seem unrelated to my changes, most likely a versioning issue due to some dependency updates (last commit on master is 3 weeks ago)

src/gt4sd/algorithms/core.py
```diff
@@ -700,6 +790,108 @@ def ensure_artifacts(self) -> str:
         return self.ensure_artifacts_for_version(self.algorithm_version)
 
 
+class ConfigurablePropertyAlgorithmConfiguration(AlgorithmConfiguration):
```
drugilsberg, line 793: wondering why all the type ignore in the methods, what is the error reported?

src/gt4sd/algorithms/core.py
```diff
@@ -720,7 +912,7 @@ def get_configuration_class_with_attributes(
 
 
 class PropertyPredictor(ABC, Generic[S, U]):
-    """WIP"""
+    """TODO: Might be deprecated in future release."""
```
drugilsberg, line 915: what about adding a `logger.warning` message in the constructor to acknowledge the deprecation?

src/gt4sd/configuration.py
```diff
@@ -53,13 +54,29 @@ class GT4SDConfiguration(BaseSettings):
...
     gt4sd_s3_host_hub: str = "s3.par01.cloud-object-storage.appdomain.cloud"
     gt4sd_s3_access_key_hub: str = "d9536662ebcf462f937efb9f58012830"
     gt4sd_s3_secret_key_hub: str = "934d1f3afdaea55ac586f6c2f729ac2ba2694bb8e975ee0b"
     gt4sd_s3_secure_hub: bool = True
-    gt4sd_s3_bucket_hub: str = "gt4sd-cos-hub-algorithms-artifacts"
+    gt4sd_s3_bucket_hub_algorithms: str = "gt4sd-cos-hub-algorithms-artifacts"
+    gt4sd_s3_bucket_hub_properties: str = "gt4sd-cos-hub-properties-artifacts"
+
+    gt4sd_s3_modules: List[str] = ["algorithms", "properties"]
+    get_local_cache_path: Dict[str, str] = {
```
drugilsberg, line 68: I would remove the get prefix from this attributes, they can be misleading and suggesting that those are methods

src/gt4sd/configuration.py
```diff
@@ -72,27 +89,33 @@ def get_instance() -> "GT4SDConfiguration":
...
 
 
-def upload_to_s3(target_filepath: str, source_filepath: str):
+def upload_to_s3(
+    target_filepath: str, source_filepath: str, module: str = "algorithms"
+):
     """Upload an algorithm in source_filepath in target_filepath on a bucket in the model hub.
     Args:
         target_filepath: path to save the objects in s3.
         source_filepath: path to the file to sync.
```
drugilsberg, line 107: update docstring

src/gt4sd/configuration.py
```diff
@@ -101,19 +124,25 @@ def upload_to_s3(target_filepath: str, source_filepath: str):
...
-def sync_algorithm_with_s3(prefix: Optional[str] = None) -> str:
+def sync_algorithm_with_s3(
+    prefix: Optional[str] = None, module: str = "algorithms"
+) -> str:
     """Sync an algorithm in the local cache using environment variables.
 
     Args:
         prefix: the relative path in the bucket (both
             on S3 and locally) to match files to download. Defaults to None.
+        module: the submodule that acts as a root for the bucket.
```
drugilsberg, line 135: default in docstring

src/gt4sd/configuration.py
```diff
@@ -142,17 +171,23 @@ def sync_algorithm_with_s3(prefix: Optional[str] = None) -> str:
     return os.path.join(folder_path, prefix) if prefix is not None else folder_path
 
 
-def get_cached_algorithm_path(prefix: Optional[str] = None) -> str:
+def get_cached_algorithm_path(
+    prefix: Optional[str] = None, module: str = "algorithms"
+) -> str:
+
+    if module not in gt4sd_configuration_instance.gt4sd_s3_modules:
+        raise KeyError(f"Unknown submodule {module}")
```
drugilsberg, line 179: Here I would raise a `ValueError` since we explicitly check for the validity of the value. A final user does not need to know this is a dictionary key.
Also I would consider expanding the error message:
```suggestion
        raise ValueError(f"Unknown cache module: {module}. Supported module: {','.join(gt4sd_configuration_instance.gt4sd_s3_modules)}")
```

src/gt4sd/configuration.py
```diff
@@ -53,13 +54,29 @@ class GT4SDConfiguration(BaseSettings):
...
 
     gt4sd_s3_host_hub: str = "s3.par01.cloud-object-storage.appdomain.cloud"
     gt4sd_s3_access_key_hub: str = "d9536662ebcf462f937efb9f58012830"
     gt4sd_s3_secret_key_hub: str = "934d1f3afdaea55ac586f6c2f729ac2ba2694bb8e975ee0b"
     gt4sd_s3_secure_hub: bool = True
-    gt4sd_s3_bucket_hub: str = "gt4sd-cos-hub-algorithms-artifacts"
+    gt4sd_s3_bucket_hub_algorithms: str = "gt4sd-cos-hub-algorithms-artifacts"
+    gt4sd_s3_bucket_hub_properties: str = "gt4sd-cos-hub-properties-artifacts"
+
+    gt4sd_s3_modules: List[str] = ["algorithms", "properties"]
```
drugilsberg, line 67: maybe a set is better:
```suggestion
    gt4sd_s3_modules: Set[str] = {"algorithms", "properties"}
```

drugilsberg: also wondering whether we need this attribute or we can simply rely on the keys of the dictionaries defined afterwards.

src/gt4sd/configuration.py
```diff
@@ -101,19 +124,25 @@ def upload_to_s3(target_filepath: str, source_filepath: str):
...
     Args:
         prefix: the relative path in the bucket (both
             on S3 and locally) to match files to download. Defaults to None.
+        module: the submodule that acts as a root for the bucket.
 
     Returns:
         str: local path using the prefix.
     """
+    if module not in gt4sd_configuration_instance.gt4sd_s3_modules:
+        raise KeyError(f"Unknown submodule {module}")
```
drugilsberg, line 141: see comment below

src/gt4sd/configuration.py
```diff
@@ -182,13 +219,16 @@ def get_algorithm_subdirectories_with_s3(prefix: Optional[str] = None) -> Set[st
     Returns:
         Set: set of available algorithms on s3 with that prefix.
     """
+    if module not in gt4sd_configuration_instance.gt4sd_s3_modules:
```
drugilsberg, line 222: see comment above

src/gt4sd/configuration.py
```diff
@@ -172,7 +207,9 @@ def get_algorithm_subdirectories_from_s3_coordinates(
     return client.list_directories(bucket=bucket, prefix=prefix)
 
 
-def get_algorithm_subdirectories_with_s3(prefix: Optional[str] = None) -> Set[str]:
+def get_algorithm_subdirectories_with_s3(
+    prefix: Optional[str] = None, module: str = "algorithms"
+) -> Set[str]:
     """Get algorithms in the s3 buckets.
 
     Args:
```
drugilsberg, line 215: update docstring

src/gt4sd/configuration.py
```diff
@@ -215,16 +255,19 @@ def get_algorithm_subdirectories_with_s3(prefix: Optional[str] = None) -> Set[st
...
 
-def get_algorithm_subdirectories_in_cache(prefix: Optional[str] = None) -> Set[str]:
+def get_algorithm_subdirectories_in_cache(
+    prefix: Optional[str] = None, module: str = "algorithms"
+) -> Set[str]:
     """Get algorithm subdirectories from the cache.
 
     Args:
         prefix: prefix matching cache subdirectories. Defaults to None.
+        module: the submodule that acts as a root for the bucket.
```
drugilsberg, line 265: default in docstring

src/gt4sd/properties/core.py
```diff
@@ -34,6 +34,21 @@ class PropertyPredictorParameters(BaseModel):
     pass
 
 
+# Base class for property predictors that use S3 artifacts
+class S3Parameters(PropertyPredictorParameters):
+
+    algorithm_type: str = "prediction"
+
+    domain: str = Field(
```
drugilsberg, line 42: consider using an enum to control better the inputs.
```
from enum import Enum

class DomainSubmodule(str, Enum):
    molecules = "molecules"
    properties = "properties"
```
And then:
```suggestion
    domain: DomainSubmodule = Field(
        ..., example=DomaineSubmodule.molecules, description="Submodule of gt4sd.properties"
     )
```
In this way if a string different from the supported ones is provided the class will fail with a validation error.

drugilsberg (changes requested):
Great job @jannisborn, an amazing PR, see my comments they should be quickly to implement and we can also discuss about them. It's mostly minor improvements in the design.

src/gt4sd/configuration.py
jannisborn: For the moment we could totally do that, but I had a usecase in mind where we add another submodule that only supports  e.g.`s3_bucket` and not `s3_bucket_hub`. As soon as we would have diverging keys in the 3 dicts, we might enter a gamble

src/gt4sd/properties/core.py
jannisborn: perfect!! I was looking for an argument like `choices` in the `Field` class that would achieve the same but I didnt find anything

src/gt4sd/algorithms/core.py
jannisborn: It's because of the parent class behavior in all cases

- So the ones in the contstructor for lines like `self.domain=domain` says: `error: Cannot assign to class variable "domain" via instance`. That's because in the parent class (`AlgorithmConfiguration`) we set it as `domain: ClassVar[str]` so its 
- the ones in the signatures like `get_application_prefix` which returns a `str` are because in the parent class those are class methods, not instance methods. THe error is `Signature of "get_application_prefix" incompatible with supertype "AlgorithmConfiguration`

It might be fixable by a refactor but I'm not sure it's worth it

jannisborn:
I closed all the trivial issues. If you agree on the two open ones, I think that we can merge. I also expanded a bit on the unittests.

jannisborn:
Just realizing, I think this PR should also close #41 @drugilsberg ?

drugilsberg:
> Just realizing, I think this PR should also close #41 @drugilsberg ?

correct 👍

src/gt4sd/algorithms/core.py
drugilsberg: Makes sense, can you track this in an issue with the improvement tag? So that we keep it on the radar.

drugilsberg (approved):
Looks great to me, we can merge.

Issue #118, opened 2022-07-20
Unify scores and property predictors


drugilsberg:
**Is your feature request related to a problem? Please describe.**
Unifying `gt4sd.domains.materials` scorers and `gt4sd.properties` seems like a natural step.

**Describe alternatives you've considered**
A good starting point can be to `gt4sd.properties` in `gt4sd.domains.materials` to define a scorer based on those.

georgosgeorgos:
* Use scores based on property_prediction to define rewards for tasks

PR #124, opened 2022-08-12
feat: property_predictors in scorer

georgosgeorgos:
* Implement `PropertyPredictorScorer` in `domains.materials.property_scorer`. 
           - circular import using `domains.materials.scorer` for the implementation 
* We are simply using the `PropertyPredictorRegistry` to select a `property` and `parameters` by `name` and `PropertyPredictionScorer` to compute a score on a `sample` wrt a `target` value.
* Tests mimick the logic in properties.


src/gt4sd/domains/materials/scorer.py
```diff
@@ -352,6 +352,15 @@ def score(self, smiles: str) -> float:
         return Chem.QED.qed(Chem.MolFromSmiles(smiles))
 
 
+SCORING_FUNCTIONS = {
+    "rdkit_scorer": RDKitDescriptorScorer,
+    "tanimoto_scorer": TanimotoScorer,
+    "isomer_scorer": IsomerScorer,
+    "smarts_scorer": SMARTSScorer,
+    "qed_scorer": QEDScorer,
+}
```
jannisborn, line 361: Maybe here we could add the remaining (more flexible) scorers, in order to have a singular place to access all the scorers

jannisborn (commented):
It looks really cool @georgosgeorgos! Let's wait for @drugilsberg opinion but I think it's almost ready to merge. Apart from a bit of cosmetics (see comments) I had three general thoughts:

- Now the `PropertyPredictorScorer` is generic in that it supports any properties, irrespective of the submodule (both `molecules` and `proteins`). But the docstring are talking only SMILES aka molecules and users might easily be fooled by thinking that all predictors available in the registry can be used by all inputs (here molecules and proteins). For the future I think it would be cleaner to have child classes like `MoleculePropertyScorer` that validate whether the constructor arguments like `name` can be used for molecules (can be done via `MOLECULE_PROPERTY_PREDICTOR_FACTORY` from `gt4sd.properties.molecules`). 
- Tangentially to that, it might even make sense to validate the incoming sequences via `Chem.MolFromSmiles` or `Chem.MolFromFASTA` (for proteins) since these scorers can be used during training where the model might collapse
- Also tangential, but many properties in `materials.scorer` are imported from Guacamol even though we have them now available in our own properties submodule

src/gt4sd/domains/materials/scorer.py
georgosgeorgos: Good point. I think I will move `SCORING_FUNCTIONS` in the init.

georgosgeorgos: If we add `PropertyPredictorScorer` we have problems with circular imports. Maybe we can solve these problems by relying more on `gt4sd.properties` in `domains`.

georgosgeorgos:
> It looks really cool @georgosgeorgos! Let's wait for @drugilsberg opinion but I think it's almost ready to merge. Apart from a bit of cosmetics (see comments) I had three general thoughts:
> 
> * Now the `PropertyPredictorScorer` is generic in that it supports any properties, irrespective of the submodule (both `molecules` and `proteins`). But the docstring are talking only SMILES aka molecules and users might easily be fooled by thinking that all predictors available in the registry can be used by all inputs (here molecules and proteins). For the future I think it would be cleaner to have child classes like `MoleculePropertyScorer` that validate whether the constructor arguments like `name` can be used for molecules (can be done via `MOLECULE_PROPERTY_PREDICTOR_FACTORY` from `gt4sd.properties.molecules`).
> * Tangentially to that, it might even make sense to validate the incoming sequences via `Chem.MolFromSmiles` or `Chem.MolFromFASTA` (for proteins) since these scorers can be used during training where the model might collapse

Thanks @jannisborn for the comments. I integrated basic validation for the property `name` and the `sample` (molecule and protein) in input. I also added child classes specific for molecule (`MoleculePropertyPredictorScorer`) and protein (`ProteinPropertyPredictorScorer`) with basic validation. 

> * Also tangential, but many properties in `materials.scorer` are imported from Guacamol even though we have them now available in our own properties submodule

We should definitely rely more on `gt4sd.properties` in `gt4sd.domains`. Importing from properties could solve some circular import issues.

jannisborn:
Cool @georgosgeorgos , it should be more robust like that!

About the circular imports: How about moving `gt4sd.domains` into `gt4sd.properties.scorers`? That might resolve the issue and tie the two modules more together (scorers are very similar to properties anyways). 
The only other thing in `domains` apart from the scorers is the `protein_encoding` but that might also fit better into `gt4sd.properties.proteins` where we could add it as `TapeEncoder` to the property factory. 
Any opinion @drugilsberg ?

src/gt4sd/domains/materials/property_scorer.py
```diff
@@ -0,0 +1,178 @@
...
+        """Scoring function that calculates a generic score for a property in molecules.
+
+        Args:
+            name: name of the property to score.
+            target: target score that will be used to get the distance to the score of a molecule or protein (not be confused with parameters["target"]).
+            parameters: parameters for the scoring function.
+        """
+        self.name = name
+        self.target = target
+        self.parameters = parameters
```
jannisborn, line 147: I might be wrong but it looks like most of these attribute assignments are redundant

jannisborn (approved):

drugilsberg:
> Cool @georgosgeorgos , it should be more robust like that!
> 
> About the circular imports: How about moving `gt4sd.domains` into `gt4sd.properties.scorers`? That might resolve the issue and tie the two modules more together (scorers are very similar to properties anyways). The only other thing in `domains` apart from the scorers is the `protein_encoding` but that might also fit better into `gt4sd.properties.proteins` where we could add it as `TapeEncoder` to the property factory. Any opinion @drugilsberg ?

On this I'm a bit hesitant as the properties are using protein/molecule representations and encoding methods, but those should not strictly live in that submodule.

georgosgeorgos:
The circular import was caused by the `{MOLECULE, PROTEIN}_PROPERTY_PREDICTOR_FACTORY` used in `PropertyPredictorScorer`. I solved these issues but not in the most elegant way: we define `SCORING_FUNCTIONS_FACTORY` in `gt4sd.properties.scores` and then we update it in `gt4sd.properties`.

jannisborn:
Nice that you found a workaround for the circular imports! It looks great to me, but at this point we could even move the PropertyPredictorScorer to gt4sd.properties.scorers (if it doesn't cause circular imports oc)

georgosgeorgos:
> Nice that you found a workaround for the circular imports! It looks great to me, but at this point we could even move the PropertyPredictorScorer to gt4sd.properties.scorers (if it doesn't cause circular imports oc)

I think we will have the same issue because we use `PropertyPredictorRegistry` and the FACTORIES.

drugilsberg (changes requested):
For me it looks great I would just move property predictor scorer somewhere else as discussed previously and decouple it from the module init, so that there we leave only the initialisation of the factories.

georgosgeorgos:
> For me it looks great I would just move property predictor scorer somewhere else as discussed previously and decouple it from the module init, so that there we leave only the initialisation of the factories.

We can do it, but then we cannot initialize the `SCORING_FUNCTIONS_FACTORY` in the init.

georgosgeorgos:
#118

src/gt4sd/properties/__init__.py
```diff
@@ -59,6 +76,41 @@ def get_property_predictor(
...
+            scorer_name
+        ]
+        return property_predictor_scorer(property_name, scoring_function, target)
+
     @staticmethod
     def list_available() -> List[str]:
         return AVAILABLE_PROPERTY_PREDICTORS
+
+    @staticmethod
+    def list_available_scorer() -> List[str]:
```
drugilsberg, line 115:
```suggestion
    def list_available_scorers() -> List[str]:
```

drugilsberg (commented):
Almost there :)

src/gt4sd/properties/scorer.py
```diff
@@ -0,0 +1,165 @@
...
+        Returns:
+            True if the input is valid.
+        """
+        # if selected property is available for molecules, check that sample is a SMILES
+        if self.name in AVAILABLE_MOLECULES_PROPERTY_PREDICTOR:
+            if Chem.MolFromSmiles(sample) is None:
+                raise ValueError(
+                    f"{property_name} is a property available for molecules and {sample} is not a valid SMILES. Please input a molecule."
+                )
+        else:
```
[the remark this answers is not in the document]
georgosgeorgos: Good point. We already perform this check in the `PropertyPredictorRegistry` but in principle the `PropertyPredictorScorer` can be used without the registry.

drugilsberg (approved):
Looks good to me, in case the CI passes will merge and release a new version.

PR #126, opened 2022-08-19
feat: configurable unverified contexts for SSL.

drugilsberg:
Signed-off-by: Matteo Manica <drugilsberg@gmail.com>

Allow to create unverified SSL contexts. This can be useful when using gt4sd properties predictors inside containers where it can be non trivial to verify local certificates to download prediction model artifacts.


jannisborn (approved):
Cool, LGTM but without an actual test it's hard to verify 😅

PR #128, opened 2022-08-23
wip: integrate diffusion models in gt4sd

georgosgeorgos:
A preliminary draft that introduces support for diffusion-based generative models in GT4SD.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,3 +24,4 @@ sacremoses>=0.0.41
 protobuf<=3.20.1
 modlamp>=4.0.0
 PyTDC>=0.3.6
+diffusers>=0.2.2
--- a/src/gt4sd/algorithms/generation/diffusion/core.py
+++ b/src/gt4sd/algorithms/generation/diffusion/core.py
@@ -1,0 +1,337 @@
+#
+# MIT License
+#
+# Copyright (c) 2022 GT4SD team
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+#
+"""HuggingFace Diffusers generation algorithm."""
+
+import logging
+from dataclasses import field
+from typing import ClassVar, Dict, Optional, Set, TypeVar
+
+from ...core import (
+    AlgorithmConfiguration,
+    GeneratorAlgorithm,
+    Untargeted,
+    get_configuration_class_with_attributes,
+)
+from ...registry import ApplicationsRegistry
+from .implementation import MODEL_TYPES, Generator
+
+logger = logging.getLogger(__name__)
+logger.addHandler(logging.NullHandler())
+
+T = type(None)
+S = TypeVar("S", bound=str)
+
+
+class HuggingFaceGenerationAlgorithm(GeneratorAlgorithm[S, T]):
+    def __init__(
+        self, configuration: AlgorithmConfiguration, target: Optional[T] = None
+    ):
+        """HuggingFace generation algorithm.
+
+        Args:
+            configuration: domain and application
+                specification, defining types and validations.
+            target: unused since it is not a conditional generator.
+
+        Example:
+            An example for using a generative algorithm from HuggingFace::
+
+                configuration = HuggingFaceXLMGenerator()
+                algorithm = HuggingFaceGenerationAlgorithm(configuration=configuration)
+                items = list(algorithm.sample(1))
+                print(items)
+        """
+
+        configuration = self.validate_configuration(configuration)
+        # TODO there might also be a validation/check on the target input
+
+        super().__init__(
+            configuration=configuration,
+            target=None,  # type:ignore
+        )
+
+    def get_generator(
+        self,
+        configuration: AlgorithmConfiguration[S, T],
+        target: Optional[T],
+    ) -> Untargeted:
+        """Get the function to sample batches.
+
+        Args:
+            configuration: helps to set up the application.
+            target: context or condition for the generation. Unused in the algorithm.
+
+        Returns:
+            callable generating a batch of items.
+        """
+        logger.info("ensure artifacts for the application are present.")
+        self.local_artifacts = configuration.ensure_artifacts()
+        implementation: Generator = configuration.get_conditional_generator(  # type: ignore
+            self.local_artifacts
+        )
+        return implementation.sample
+
+    def validate_configuration(
+        self, configuration: AlgorithmConfiguration
+    ) -> AlgorithmConfiguration:
+        # TODO raise InvalidAlgorithmConfiguration
+        assert isinstance(configuration, AlgorithmConfiguration)
+        return configuration
+
+
+@ApplicationsRegistry.register_algorithm_application(HuggingFaceGenerationAlgorithm)
+class HuggingFaceConfiguration(AlgorithmConfiguration[str, None]):
+    """Basic configuration for an hugging face algorithm."""
+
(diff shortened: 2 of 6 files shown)


georgosgeorgos:
Diffusion models for molecules: https://github.com/huggingface/diffusers/pull/54

drugilsberg (commented):
Looks good to me, I wonder whether it makes sense to also add a training pipeline to train/fine-tune your own diffusion models.

georgosgeorgos:
> Looks good to me, I wonder whether it makes sense to also add a training pipeline to train/fine-tune your own diffusion models.

Good point. Working on it.

georgosgeorgos:
TODO:

- [ ] add notebook

jannisborn:
Created a PR to fix [da05f92](https://github.com/GT4SD/gt4sd-core/pull/128/commits/da05f9225307973f517835e23a47913e51212049) in https://github.com/georgosgeorgos/gt4sd-core/pull/1

drugilsberg (changes requested):
Mostly comments on the training pipelines.

src/gt4sd/training_pipelines/diffusion/core.py
```diff
@@ -0,0 +1,424 @@
...
+logger = logging.getLogger(__name__)
+logger.addHandler(logging.NullHandler())
+
+# NOTE: This import is needed because importing torchvision before rdkit Descriptors
+# can cause segmentation faults.
+Descriptors
+
+
+class DiffusionTrainingPipeline(TrainingPipeline):
+    """Diffusion training pipelines."""
```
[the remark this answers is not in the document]
georgosgeorgos: Still waiting for the integration of diffusion models on molecules in diffusers. It will be probably easier to have two separate TrainingPipeline. What about `DiffusionTrainingPipelineVision`?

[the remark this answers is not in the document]
drugilsberg: Not bad, but to be consistent with the other pipelines it would be cool to keep as suffix `TrainingPipeline`.
Maybe `DiffusionForVisionTrainingPipeline`? or something along this line

georgosgeorgos:
I'm using conditional requirements for the CI as suggested by @drugilsberg. If everything works, we can get rid of the `requirements_ci.txt` and `conda_ci.yml` and simplify the pipeline a bit.

requirements.txt
```diff
@@ -1,13 +1,14 @@
...
+diffusers>=0.2.4
 joblib>=1.1.0
 keras==2.3.1
-torchmetrics<0.7
 keybert==0.2.0
 minio==7.0.1
 modlamp>=4.0.0
-scikit-learn<0.24.0
 numpy>=1.16.5
+protobuf<3.20
```
jannisborn, line 11: Thanks for this, I used `sort-requirements` to sort them lately and I have really no clue how I ended up failing on sorting this file 😱

src/gt4sd/algorithms/generation/diffusion/core.py
```diff
@@ -0,0 +1,353 @@
...
+    )
+    prefix: str = field(
+        default="",
+        metadata=dict(
+            description="Text defining context provided prior to the prompt."
+        ),
+    )
+    number_of_sequences: int = field(
+        default=8,
+        metadata=dict(description="Number of text sequences to generate."),
```
jannisborn, line 160: does this refer only to text models or is this just `number_of_samples`?

jannisborn (approved):
Fantastic work, really enjoyable to review @georgosgeorgos 

Just a few thoughts/cpmments
- I see this PR is a general wrapper for diffusers but since most of their model currently are text to image, having it in `conditional_generation` instead of `generation` could make more sense?
- On the same note: Currently `domain` is set to "generic" in all cases. But instead we have modality which seems to be a concept from Diffusers. Why not overwriting the domain attribute in all child classes of `DiffusersConfiguration` - each algorithm has a specific domain, no? That would be more consistent and it could help us in the future to list algorithms selectively, e.g., by running `list_available(domain='images')`
- by convention, the forward pass in the other inference algorithm occurs via `model.sample(n)` whereas here we use `model.generator(prompt=prompt)`. Maybe unifying this would be more consistent even though we cant have a `n` > 1 for Diffusers, I guess (are the models deterministic? If not, one could and it could even make sense to ask for multiple images per prompt)
- Since images cannot be serialized trivially, the current API will throw an error when text-to-image diffusers are used. We discussed whether this should be addressed in the core and we decided to go for a `sample_serialize` that works under the hood. This PR could incentivize us to move forward on that front soon.

src/gt4sd/algorithms/generation/diffusion/core.py
```diff
@@ -0,0 +1,353 @@
...
+
+    def validate_configuration(
+        self, configuration: AlgorithmConfiguration
+    ) -> AlgorithmConfiguration:
+        assert isinstance(configuration, AlgorithmConfiguration)
+        return configuration
+
+
+@ApplicationsRegistry.register_algorithm_application(DiffusersGenerationAlgorithm)
+class DiffusersConfiguration(AlgorithmConfiguration[str, None]):
```
drugilsberg, line 103: I have the feeling some parameters are coming right away from the HuggingFace wrappers. A lot of them are unused in the implementation.py, e.g., repetition_penalty.
Let's try to clean those up and also to capture what we pass to the sample method and parameters of the configuration.

drugilsberg (changes requested):

src/gt4sd/algorithms/generation/diffusion/core.py
georgosgeorgos: this is just for text. I remove it for the moment.

georgosgeorgos:
> * I see this PR is a general wrapper for diffusers but since most of their model currently are text to image, having it in `conditional_generation` instead of `generation` could make more sense?

We should consider it. In general, we can use the models unconditionally or conditionally. But the most common use case is for sure conditional generation.

> * On the same note: Currently `domain` is set to "generic" in all cases. But instead we have modality which seems to be a concept from Diffusers. Why not overwriting the domain attribute in all child classes of `DiffusersConfiguration` - each algorithm has a specific domain, no? That would be more consistent and it could help us in the future to list algorithms selectively, e.g., by running `list_available(domain='images')`

Now the domain is `vision`.

> * by convention, the forward pass in the other inference algorithm occurs via `model.sample(n)` whereas here we use `model.generator(prompt=prompt)`. Maybe unifying this would be more consistent even though we cant have a `n` > 1 for Diffusers, I guess (are the models deterministic? If not, one could and it could even make sense to ask for multiple images per prompt)

We fixed this. Now the interface is standardized.

> * Since images cannot be serialized trivially, the current API will throw an error when text-to-image diffusers are used. We discussed whether this should be addressed in the core and we decided to go for a `sample_serialize` that works under the hood. This PR could incentivize us to move forward on that front soon.

I had to use a workaround for the sampling method in `gt4sd/algorithms/core.py`. For the moment it works and we can sample images or text.

georgosgeorgos:
> I'm using conditional requirements for the CI as suggested by @drugilsberg. If everything works, we can get rid of the `requirements_ci.txt` and `conda_ci.yml` and simplify the pipeline a bit.

`Conditional import` works for the CI!

drugilsberg (commented):
Suggested a simplification. The rest look good to me, but I would like to know what do you think (@georgosgeorgos) and see the test outcome.

georgosgeorgos:
> Suggested a simplification. The rest look good to me, but I would like to know what do you think (@georgosgeorgos) and see the test outcome.

Great improvement!

drugilsberg (approved):

src/gt4sd/algorithms/generation/diffusion/core.py
```diff
@@ -0,0 +1,325 @@
...
+
+    def get_target_description(self) -> Optional[Dict[str, str]]:
+        """Get description of the target for generation.
+
+        Returns:
+            target description, returns None in case no target is used.
+        """
+        return None
+
+    def get_conditional_generator(
```
drugilsberg, line 146: Here we are breaking the signature by adding target. To keep the implementation consistent with the other algorithms we should remove target and passing it via the configuration since we placed the diffusion algorithm in the generation submodule.
The same pattern is followed from the HuggingFace wrappers.

drugilsberg (changes requested):
I realised we need to adjust the handling of the target if we want to follow the library pattern.

jannisborn (changes requested):
> > * I see this PR is a general wrapper for diffusers but since most of their model currently are text to image, having it in `conditional_generation` instead of `generation` could make more sense?
> 
> We should consider it. In general, we can use the models unconditionally or conditionally. But the most common use case is for sure conditional generation.

In this case I think that we should move it to `conditional_generation` in oder to not further confound the differences between both submodules

src/gt4sd/algorithms/generation/diffusion/implementation.py
```diff
@@ -0,0 +1,162 @@
...
+            self.model = model_class.from_pretrained(
+                model_name_or_path,
+                use_auth_token=self.auth_token,
+            )
+        else:
+            self.model = model_class.from_pretrained(model_name_or_path)
+        self.model.to(self.device)
+
+    def sample(
+        self, batch_size: int = 1, prompt: Union[str, List[str]] = None
```
drugilsberg, line 144: Two comments:
- why not removing the prompt at this point? no strong feelings, since the implementation is hidden
- we should rename batch_size to number of samples, what do you think?

drugilsberg (commented):
I think we are almost ready.
I understand @jannisborn comment, but I think that sticking to what we did for the HuggingFace wrapper is a sensible solution. We could also refactor later.

src/gt4sd/algorithms/generation/diffusion/implementation.py
georgosgeorgos: Good points. Done.

drugilsberg (approved):
Looks good to me for an initial release.

PR #129, opened 2022-08-25
fix: missing deps in pip installation

jannisborn:
- adding missing requirements (`loguru` and `joblib` are used inside the LM trainer)
- syncing requirements.txt and setup.cfg

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,3 +24,5 @@ sacremoses>=0.0.41
 protobuf<=3.20.1
 modlamp>=4.0.0
 PyTDC>=0.3.6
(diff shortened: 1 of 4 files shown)


drugilsberg (changes requested):
@jannisborn can we get rid of the loguru dependency and use logging with the NullHandler as we do everywhere else. Loguru makes handling the logs very difficult if used outside of a CLI in a library.

jannisborn:
Done!

drugilsberg (approved):
Great job, as soon as CI passes will merge and release the new version.

PR #131, opened 2022-09-01
Fix the link to the CLA in the issue template

avaucher:
The current link ends up resolving to https://github.com/GT4SD/CONTRIBUTING.md#contributor-license-agreement, which leads to a "Page not found".


drugilsberg (approved):
Thanks for catching this 🙏

Issue #92, opened 2022-06-20
Problem multiprocess in requirements


georgosgeorgos:
The new multiprocess library version (`0.70.13`) gives problems when installing `gt4sd-core` using the development mode. I had to set `multiprocess==0.70.12.2` to install the library.

drugilsberg:
Thanks for reporting.
Here it would be great to have a stack/error trace to understand what errors exactly.
I tried a build both on Ubuntu and Mac OS X and it went through without pinning the requirements.

georgosgeorgos:
Installing in development mode `pip install --no-deps -e .` we have a problem related to the `wheel` version:

gt4sd-core git:(main) ✗ gt4sd-inference --algorithm_name PaccMannRL --algorithm_application PaccMannRLProteinBasedGenerator --target MVLSPADKTNVKAAWGKVGAHAGEYGAEALERMFLSFPTT --number_of_samples 10

Traceback (most recent call last):
  File "/opt/anaconda3/envs/gt4sd_tmp/lib/python3.7/site-packages/pkg_resources/__init__.py", line 579, in _build_master
    ws.require(__requires__)
  File "/opt/anaconda3/envs/gt4sd_tmp/lib/python3.7/site-packages/pkg_resources/__init__.py", line 897, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/opt/anaconda3/envs/gt4sd_tmp/lib/python3.7/site-packages/pkg_resources/__init__.py", line 788, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (protobuf 3.20.1 (/opt/anaconda3/envs/gt4sd_tmp/lib/python3.7/site-packages), Requirement.parse('protobuf<3.20,>=3.9.2'), {'tensorboard'})

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/anaconda3/envs/gt4sd_tmp/bin/gt4sd-inference", line 33, in <module>
    sys.exit(load_entry_point('gt4sd', 'console_scripts', 'gt4sd-inference')())
  File "/opt/anaconda3/envs/gt4sd_tmp/bin/gt4sd-inference", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/opt/anaconda3/envs/gt4sd_tmp/lib/python3.7/site-packages/importlib_metadata/__init__.py", line 203, in load
    module = import_module(match.group('module'))
[...]

PR #134, opened 2022-09-06
fix: update requirements to set up CLI commands

jannisborn:
Solves to #92 which was closed even though those changes were never fixed on `main`.

Effectively in the dev setup thus far, the CLI commands were not functional since `wheel` and `protobuf` were complaining


drugilsberg (approved):
It looks good to me, I thought the issue was closed because the problem was not relevant anymore.

jannisborn:
I also took the opportunity to sort all requirements we have alphabetically. For the future we should add new requirements in the correct line

jannisborn:
I also added an example (to be expanded) under `examples/CLI/README.md` which lists all the available algorithms for inference.
I think it can help to get started and get an overview of all inference algorithms that are there

jannisborn:
Closes #92

drugilsberg (approved):
Amazing work

PR #140, opened 2022-09-25
reqs: gpu requirements

georgosgeorgos:
* Separate GPU requirements
* Use CPU requirements for CI


drugilsberg (changes requested):
Great idea, it's just a bit difficult to maintain, why not having:
- requirements.txt: containing all shared deps
- cpu_requirements.txt: containing the conditional block installing the cpu versions
- gpu_requirements.txt: containing the gpu versions
Then it's just about updating the two conda.yml files.

drugilsberg (approved):
Looks great :)

PR #145, opened 2022-10-04
TDC Oracle integration

### LINKED ISSUE #117: TDC oracles based on API keys
Currently we have no mechanism to use TDC oracles based on API keys, we should evaluate whether this is needed.

### LINKED ISSUE #119: Relax sklearn version dependency caused by TDC
Some of our property computations rely on TDC and thus an old and specific sklearn version

Upstream solution by TDC would be ideal: https://github.com/mims-harvard/TDC/issues/163 either with 2 COS objects and an if/else parsing of the package version or a cheap fix that emulates the presence of `sklearn.svm.classes` in more recent sklearn version

jannisborn:
Closes #119 
Closes #117 

- bumping tdc dependency which finally allows to relax the sklearn version
- support for three more complex oracles.  For details see [here](https://tdcommons.ai/functions/oracles/).
  - Two of them (ASKCOS and MoleculeOne) give synthesizability scores similar to RXN retroworkers. They need an api key or a the IP of the API endpoint to run. I created schemata to handle those 
  - Third run is for docking. I simply populated the TDC interface that either allows to dock against an existing target (that is fetched automatically from their server) or against a user-provided target. In both cases, however, the user needs to install a lot of software manually before these properties work
  - No unit tests provided in all cases, since installation is either manual or requires authentification

- I had to temporarily pin `importlib_metadata` since they just released v5 which comes with some errors, see [here](https://stackoverflow.com/questions/73929564/entrypoints-object-has-no-attribute-get-digital-ocean), original issue [here](https://github.com/python/importlib_metadata/issues/409)
- Unrelated: I fixed the legacy values of `test_fraction` in RT Trainer code
- Unrelated: Renamed `SIDER` property to `Sider` in order to adhere to our styling

### DIFF
--- a/examples/regression_transformer/README.md
+++ b/examples/regression_transformer/README.md
@@ -11,7 +11,7 @@ conda activate gt4sd
 To launch a finetuning of a RT pretrained on drug-like moelcules from ChEMBL, execute the following from the GT4SD root:
 
 ```console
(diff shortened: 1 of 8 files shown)


drugilsberg (approved):
Great job @jannisborn, I restarted the CI to make sure the failed test is due to some stochastic behaviour in one of the TorchDrug tests.

jannisborn:
My bad, I broke the test of the RT training pipeline when removing the legacy argument. It should work now

PR #148, opened 2022-10-06
fix: change default to true

georgosgeorgos:
`transformers==4.22.0` changed `parse_json_file(self, json_file: str, allow_extra_keys: bool = False)`, with allow_extra_keys=False as default and raising an error. We always use extra_keys in the library and this change breaks the training pipeline.


drugilsberg (approved):
Good catch, this looks good to me.

christofid (approved):
Looks good to me!

Issue #156, opened 2022-10-19
RT: gt4sd-trainer pipeline falls back to HF hub when the starting model is not in the local cache


drugilsberg:
**Describe the bug**

When running `gt4sd-trainer` using a starting model not present in the local cache, there is fallback mechanism to look for the model in the HF hub.

**To Reproduce**
1. Install gt4sd or simply use the base docker image
2. Run the example for RT training

**Expected behavior**
The command should trigger a sync of the remote cache.

jannisborn:
Should be solved via #157

PR #157, opened 2022-10-19
Tolerance values RegressionTransformer

jannisborn:
- Expanding controllability of the tolerances by setting one tolerance value per property. So tolerance can be provided as `float` or now also as `Dict[str, float]`. Missing keys are imputed with 20% and additional keys are ignored. Warnings are raised. Tests are implemented.
Example:
```py
target = 'C1(=O)O[C@@H](C)C(=O)O[C@H]1C.C2CC(NC(NC1=CC=C(OC)C=C1)=S)CCC2'
    algorithm_version='logp_and_synthesizability', search='sample', tolerance={'<logp>': 20, '<scs>': 10},
    sampling_wrapper={
        'property_goal': {'<logp>': 0.15, '<scs>': 3.5}, 'fraction_to_mask': 0.4, 'tokens_to_mask': ['C']
    }
)
)
```
- I found a bug in my previous tolerance implementation, basically the provided tolerance was not passed down and we always used a tolerance of 20%... My bad. It's fixed now.
- Clarified the README about RT finetuning, stating that the model has to be cached locally for this to work. This relates to #156 but I'm not sure we have to trigger syncing from HF model hub about RT models. None of the RT models is available in HF model hub.

Housekeeping:
- Support new version of `diffusers` (0.6.0) while maintaining support for older ones

### DIFF
--- a/src/gt4sd/algorithms/generation/diffusion/implementation.py
+++ b/src/gt4sd/algorithms/generation/diffusion/implementation.py
@@ -26,11 +26,11 @@ Implementation details for huggingface diffusers generation algorithms.
 
 Parts of the implementation inspired by: https://github.com/huggingface/diffusers/blob/main/examples/train_unconditional.py.
 """
-
 import logging
 import os
 from typing import Any, List, Optional, Union
 
+import importlib_metadata
 import numpy as np
 import torch
 from diffusers import (
@@ -45,8 +45,13 @@ from diffusers import (
     ScoreSdeVeScheduler,
     StableDiffusionPipeline,
 )
+from packaging import version
 
 from ....frameworks.torch import device_claim
+
+OLD_DIFFUSERS = version.parse(importlib_metadata.version("diffusers")) < version.parse(
+    "0.6.0"
+)
 
 logger = logging.getLogger(__name__)
 logger.addHandler(logging.NullHandler())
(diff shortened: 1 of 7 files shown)


src/gt4sd/algorithms/generation/diffusion/implementation.py
```diff
@@ -45,9 +45,14 @@
     ScoreSdeVeScheduler,
     StableDiffusionPipeline,
 )
+from packaging import version
 
 from ....frameworks.torch import device_claim
 
+OLD_DIFFUSERS = version.parse(importlib_metadata.version("diffusers")) < version.parse(
```
drugilsberg, line 52: May I suggest `DIFFUSERS_VERSION_LT_0_6_0`?
```suggestion
DIFFUSERS_VERSION_LT_0_6_0 = version.parse(importlib_metadata.version("diffusers")) < version.parse(
```

src/gt4sd/algorithms/generation/diffusion/implementation.py
```diff
@@ -153,5 +158,9 @@ def sample(self, number_samples: int = 1) -> List[Any]:
             item = self.model(batch_size=number_samples, prompt=self.prompt)
         else:
             item = self.model(batch_size=number_samples)
-        item = item["sample"]
+        # To support old diffusers versions (<0.6.0)
+        if OLD_DIFFUSERS:
```
drugilsberg, line 162:
```suggestion
        if DIFFUSERS_VERSION_LT_0_6_0:
```

drugilsberg (commented):
Amazing job, I only left a small comment on a variable rename that I think makes more explicit and more adaptable the solution in case of other version bounds on `diffusers` will be used.

drugilsberg (approved):
LGTM

PR #163, opened 2022-11-02
RT: Adding support for explicit substructure control during generation

### LINKED ISSUE #139: RT sampling_wrapper to specify a substructure or series of tokens to keep unmasked
I would like to propose an upgrade on the feature demonstrated in this notebook:
https://github.com/GT4SD/gt4sd-core/blob/main/notebooks/regression-transformer-demo.ipynb (see cells 12-14)

In addition to explicitly specifying tokens_to_mask, one probably could more likely imagine that a chemist might want to specify a substructure to mask or to "freeze" (keep unchanged, i.e. unmasked). It might be easier to specify tokens to freeze as that would be just selecting a part of the string to be kept unmasked. Prototype example is given below.

```
    sampling_wrapper={
        'property_goal': {
            '<logp>': 6.123,
            '<scs>': 1.5
        },
        'fraction_to_mask': 0.6,
        # keep morpholino tail unchanged
        'tokens_to_freeze': ['N4CCOCC4']
    }
```
If one could specify substructure to freeze or to mask - that would be potentially even more advantageous, as that would remove ambiguities when a substructure can be expressed in more than one sequence.

```
    sampling_wrapper={
        'property_goal': {
            '<logp>': 6.123,
            '<scs>': 1.5
        },
        'fraction_to_mask': 0.6,
        # keep morpholino tail unchanged
        'substructure_to_freeze': ['N1CCOCC1'],
        # explicitly mask benzene ring moiety
        'substructure_to_mask':  ['C1=CC=CC=C1'],
    }
```

One could use RDKit functionality to identify substructure tokens, as given here: https://www.rdkit.org/docs/Cookbook.html#substructure-matching

Regarding the interpretation of the 'fraction_to_mask', I would then imagine that it would best applied to the remaining set of tokens (after tokens_to_freeze and explicit tokens_to_mask are excluded).
I hope this makes sense, happy to clarify and exemplify further.

jannisborn:
Closes #139 

This issue enables support is an extension of the `sampling_wrapper` for more controllable conditional generation in the RT.
Two new arguments are supported:
- `substructures_to_keep`: A list of substructures that are excluded from masking
- `substructures_to_mask`: A list of substructures that are definitely masked

In both cases, **all** occurrences of those substructures are masked. The behavior of `fraction_to_mask` and `tokens_to_mask` does not change but now this stochastic masking only applies on tokens that are not explicitly controlled.

Note that a fundamental assumption is that the provided substructures can be matched in the SELFIES sequence.

Example usage:
```py
from gt4sd.algorithms.conditional_generation.regression_transformer import (
    RegressionTransformer, RegressionTransformerMolecules, RegressionTransformerProteins
)
target = "CC(C#C)N(C)C(=O)NC1=CC=C(Cl)C=C1"
keep = "C1=CC=C(Cl)C=C1"
mask = 'C(=O)'

    algorithm_version="qed",
    search="sample", # the alternative is 'greedy' but 'sample' is recommended for generative tasks
    temperature=1, 
    tolerance=100, # percentage of tolerated deviation from the desired property value (here -3.53)
    sampling_wrapper={
        'property_goal': {'<qed>': 0.5}, 
        'fraction_to_mask': 0.9,
        'substructures_to_keep': [keep],
        'substructures_to_mask': [mask],
        'text_filtering': False
    }
)
)
generations = list(generator.sample(8))
generations

```

We apply post-hoc filtering to ensure the provided substructures are present. If the matching cant be done in `rdkit` the matching can optionally be done on a string level (that's very crude but might be better than nothing). It's disabled per default but can be set with `text_filtering` argument.


drugilsberg (approved):
Great job 👏

PR #164, opened 2022-11-04
Minors

### LINKED ISSUE #133: Remove pydantic upper bound
Make pydantic compatible with versions >=1.9.2
Currently newer pydantic versions appear to be incompatible with registry implementation raising this error when calling `functools.update_wrapper`:

```console
AttributeError: 'DataclassProxy' object has no attribute 'algorithm_type'
```

See here pydantic release notes where they refer to a consistent refactor of `dataclass`: https://github.com/pydantic/pydantic/releases/tag/v1.10.0.

### LINKED ISSUE #160: Install MoLeR  from PyPI
Install MoLeR  from PyPI since it became available (0.0.3)

**Problem**
Currently building the conda env succeeds, but importing torchdrug creates segfaults on both macOS and ubuntu machines

jannisborn:
- move moler out from vcs requirements (closes #160)
- fix problem with pydantic upper bound (closes #133). I opened an issue with pydantic [here](https://github.com/pydantic/pydantic/issues/4695) and our current fix is more a hotfix.
- fixed a problem introduced in the GeoDiff PR that introduced `nglviewer` as dependency. When installing jupyter now, nglviewer raises an issue. I looked into that, it can be fixed by upper-bounding `ipywidgets` which is now done in `nobteooks/requirements.txt`
- bump TdC to the freshly released 0.3.8 to avoid the bug in the MoleculeOne retro wrapper


jannisborn:
Pinning pytdc for now, new issue coming up when bumping to 0.3.8

jannisborn:
Upper-bounding diffusers for the moment. 0.7.0 was released yesterday but even the builds on their main branch are failing so let's wait for their stable release.

requirements.txt
```diff
@@ -1,7 +1,7 @@
 # pypi requirements
 accelerate>=0.12
 datasets>=1.11.0
-diffusers>=0.6.0
+diffusers<=0.6.0
```
drugilsberg, line 4: is this upper bound really needed?

jannisborn: Temporarily yes, `diffusers` main builds are failing as I wrote above. See here: https://github.com/huggingface/diffusers

jannisborn: Also here I fixed the thing that we actually do support diffusers < 0.6.0

src/gt4sd/algorithms/registry.py
```diff
@@ -186,6 +186,9 @@ def decorator(
                 ],  # type: ignore
             )
 
+            PydanticConfiguration: Type[AlgorithmConfiguration] = dataclass(  # type: ignore
```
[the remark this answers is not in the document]
jannisborn: Check out the issue I opened on pydantic, it's linked above. The method has side effects, so calling it twice goes around the problem described in my issue

drugilsberg (approved):

Issue #143, opened 2022-10-04
Support in RT Trainer for multiple entities


jannisborn:
**Is your feature request related to a problem? Please describe.**
Expand Regression Transformer trainer to support multi-entity discriminations, i.e., support the `multientity_cg` collator from the RT repo

**Describe the solution you'd like**
Passing additional arguments of this collator as command line arguments

**Describe alternatives you've considered**
Reading the three new parameters from the json, not preferred since other parameters are passed from command line

PR #166, opened 2022-11-07
feat: Support in RT Trainer for multiple entities

NicolaiRee:
Solving #143 by expanding the Regression Transformer trainer to support multi-entity discriminations, i.e., support the multientity_cg collator from the RT repo.

Signed-off-by: Nicolai Ree <nicolairee@hotmail.com>


NicolaiRee:
I cannot assign reviewers for this PR, but now both checks have finally passed 🎊🥳🍾

jannisborn (commented):
Great stuff 🚀 
It looks great to me, I think all we need before we can merge is a unit test of this functionality. 

I can take care of that though

jannisborn (approved):
LGTM

PR #172, opened 2022-12-07
implementing a finetuning feature

mirunacrt:
Signed-off-by: Miruna Cretu <miruna.t.cretu@gmail.com>


jannisborn (changes requested):
Great job @mirunacrt! Just minor comments, I would call the argument `checkpoint_path` instead of `train_from` and then swap the two commands in the if clause such that the logging message is only printed if the restoration was successful. 

Then, you need to sign the CLA to trigger the tests. For that just add your mail above to the mail addresses in your github account, then open an issue on this repo by using the "CLA Signature" template. This should trigger the CI pipeline.

jannisborn (approved):

PR #178, opened 2022-12-16
Moses upgrade

### LINKED ISSUE #176: MOSES VAE from Guacamol training reconstruction is "incorrect"
**Describe the bug**
The [VAE in GT4SD](https://github.com/GT4SD/gt4sd-core/blob/main/src/gt4sd/training_pipelines/moses/vae/core.py) uses the wrapper of the Moses VAE from [Guacamol](https://github.com/BenevolentAI/guacamol_baselines/blob/master/moses_baselines/vae_train.py). Unfortunately, the decoding training step from the [Moses VAE](https://github.com/molecularsets/moses/blob/master/moses/vae/trainer.py) is bugged.
 
**More detail**
The problem arises from the definition of the `forward_decoder` [method](https://github.com/molecularsets/moses/blob/master/moses/vae/model.py#L134):
```python
def forward_decoder(self, x, z):
    lengths = [len(i_x) for i_x in x]

    x = nn.utils.rnn.pad_sequence(x, batch_first=True, padding_value=self.pad)
    x_emb = self.x_emb(x)

    z_0 = z.unsqueeze(1).repeat(1, x_emb.size(1), 1)
    x_input = torch.cat([x_emb, z_0], dim=-1)  # <--- PROBLEM 1
    x_input = nn.utils.rnn.pack_padded_sequence(x_input, lengths, batch_first=True)

    h_0 = self.decoder_lat(z)
    h_0 = h_0.unsqueeze(0).repeat(self.decoder_rnn.num_layers, 1, 1)

    output, _ = self.decoder_rnn(x_input, h_0)

    output, _ = nn.utils.rnn.pad_packed_sequence(output, batch_first=True)
    y = self.decoder_fc(output)

    recon_loss = F.cross_entropy(  # <--- PROBLEM 2
        y[:, :-1].contiguous().view(-1, y.size(-1)),
        x[:, 1:].contiguous().view(-1),
        ignore_index=self.pad
    )

    return recon_loss
```

Namely, the reconstruction step is wrong in two spots:
1. construction of the _true_ input: `x_input = torch.cat([x_emb, z_0], dim=-1)`
In the visual representation of a typical RNN, the true token feeds in from the 'bottom" of the cell and the previous hidden state from the "left". In this implementation, the reparameterized latent vector `z` is fed in both from the "left" (normal) and the "bottom" (atypical).
**Fix**: this line should be removed
2. calculation of the reconstruction loss: `recon_loss = F.cross_entropy(...)`
This reconstruction loss is calculated as the _per-token_ loss of the input batch (i.e., the mean of a batch of tokens) because the [default reduction](https://pytorch.org/docs/stable/generated/torch.nn.functional.cross_entropy.html#torch.nn.functional.cross_entropy) in `F.cross_entropy` is `"mean"`. In turn, this results in reconstruction losses that are _very_ low for the VAE, causing the optimizer to ignore the decoder and focus on the encoder. When a VAE focuses too hard on the encoder, you get mode collapse, and that's what happens with the Moses VAE.
**Fix**: this line should be: `F.cross_entropy(..., reduction="sum") / len(x)`

**To reproduce**
1. Problem 1 is not a "problem" so much as it is _highly_ atypical to structure a VAE like this. I can't say if it results in any actual problems, but it simply shouldn't be there
2. Problem 2 can be observed with two experiments:
    1. Using PCA with two dimensions, plot the embeddings of a random batch `z ~ q(z|x)` and a sample from the standard normal distribution `z ~ N(0, I)`. The embeddings from the encoder will look like a point at (0, 0) compared to the samples from the standard normal
    2. Measure the reconstruction accuracy `x_r ~ p(x | z ~ q(z | x_0))`. In a well-trained VAE, `sum(x_r == x_0 for x_0 in xs) / len(xs)` should be above 50%. This VAE is generally fairly low (in my experience).

jannisborn:
- Bumping `moses` to v0.1.1 to fix #176 (some fixes in the VAE decoder)
- Equipping all moses/Guacamol generators with the ability to filter out invalid generations
- Equip the moses VAE trainer with the ability to warm-start a model
- Bumping `guacamol_baselines` to v0.0.3 to allow model warm-starting
- Update discovery demo with a means to use the PaccMannVAE in an unbiased way. This will make it behave similarly to the Moses VAE


drugilsberg (approved):
LGTM, once the build passes we can merge

jannisborn:
I think the build will fail since the fix in moses is breaking the inference pipeline of the existing model -- I'll restart the build once I updated the model on the COS

PR #182, opened 2023-01-06
Adapting discovery demo for solubility task

jannisborn:
- Solving #181: Improving PaccMannGP to support arbitrary callables as suggested in the discussion
- Adding a new model `PaccMannVAE` that produces unconditional molecules. It's the same decoder as in PaccMannRL but just used by sampling randomly. We basically just wrap the PaccMannRL model.
- Adapting the case study on GENTRL's DDR1-inhibitor to optimize solubility instead of QED. Reran all the models, including the new one (PaccMannVAE). Findings are the same in that MoLeR and RT are performing the best


drugilsberg (approved):
The logic looks good, I did not have a chance to run the full notebook, but the logic and the updates look good.

drugilsberg:
@jannisborn small build due to a PaccMannGP test:

```console
if self.mode == "max":
[5965](https://github.com/GT4SD/gt4sd-core/actions/runs/3853838762/jobs/6567211532#step:9:5966)
            return 1 - (sum(values) / len(values))
[5966](https://github.com/GT4SD/gt4sd-core/actions/runs/3853838762/jobs/6567211532#step:9:5967)
        elif self.mode == "min":
[5967](https://github.com/GT4SD/gt4sd-core/actions/runs/3853838762/jobs/6567211532#step:9:5968)
>           return sum(values) / len(values)
[5968](https://github.com/GT4SD/gt4sd-core/actions/runs/3853838762/jobs/6567211532#step:9:5969)
E           ZeroDivisionError: division by zero
```

Looks like it is something dependent on the error handling in the scoring function in the package.

Issue #186, opened 2023-01-20
CGCNN


jannisborn:
**Describe the bug**
1. The predictions vary dependent on the total number of samples in the batch. I cant find the reason for this, but it does not occur in the original data/repo
2. The model does not reproduce the results from the public repo: https://github.com/txie-93/cgcnn

**To Reproduce**

```py
from gt4sd.properties.crystals.core import BandGapParameters, BandGap
model_parameters = BandGapParameters(algorithm_version="v0")
model = BandGap(model_parameters)
model(input="cgcnn-sample")
```
returns:
```txt
{'cif_ids': ['9011998',
  '9011050',
  '9000046',
  '9012304',
  '1101051',
  '1507756',
  '9009743',
  '1000050',
  '7206075',
  '1000041'],
 'predictions': [2.083367347717285,
  1.6597893238067627,
  1.2090169191360474,
  4.408855438232422,
  0.5542303323745728,
  0.9335317611694336,
  3.4029479026794434,
  2.7800779342651367,
  2.1562294960021973,
  4.579179286956787]}
```

Problem 1:
When simply removing the first line of the label file (`id_prop.csv`), the result is:
```txt
{'cif_ids': ['9011050',
  '1000050',
  '9000046',
  '9009743',
  '1101051',
  '1507756',
  '9011998',
  '7206075',
  '1000041'],
 'predictions': [1.6771466732025146,
  3.372532606124878,
  3.0985941886901855,
  3.7642505168914795,
  0.5874173641204834,
  0.7363685369491577,
  2.2318031787872314,
  1.8756965398788452,
  5.089865684509277]}
```
While the computation is deterministic and the order is different, it differs from the previous results with one more sample in the batch. I did verify that this **not** an issue in the original codebase.

Problem 2: In either of the cases, the result does not align with the results I obtain from the original code:
```csv
PREDICTION LABEL ID
2.0028774738311768 10.0 9012304
4.981688499450684 7.0 9009743
0.9553935527801514 9.0 9011998
1.9109779596328735 4.0 1507756
2.6892075538635254 5.0 7206075
4.7740559577941895 1.0 1000041
0.9553935527801514 9.0 9011998
4.7740559577941895 1.0 1000041
4.7740559577941895 1.0 1000041
1.5325593948364258 8.0 9011050
2.0292701721191406 3.0 1101051
1.7813189029693604 6.0 9000046
4.381993770599365 2.0 1000050
```

christofid:
[...]

PR #187, opened 2023-01-23
CGCNN Refactoring

jannisborn:
- Refactored CGCNN models such that they can be executed only with `.cif` files. Before it was needed to also provide a `json` file with the node initializaiton and a `.csv` file with the (dummy) labels. See [notebooks/cgcnn-sample](https://github.com/gt4sd/gt4sd-core/tree/main/notebooks/cgcnn-sample) for an example
- Added tests for CGCNN property prediction (has to be enabled, after #186 is fixed)
- enabled tests for the CGCNN trainer (before it was skipped in the CI. It required some modifications, but now the tests work)

Let's keep this a draft PR until #186 is fixed


christofid (approved):
Looks good to me

PR #188, opened 2023-01-25
Updating pymatgen installation/deps

drugilsberg:


jannisborn:
Tests are failing due to a legacy keras import, I quickly checked, I cant reproduce the error locally even when upgrading my deps

jannisborn (approved):
build fixed

PR #190, opened 2023-01-28
test: enable cgcnn test

jannisborn:
Even though the tests were active, they were extremely lenient since `atol=10`. Fixed in this PR


drugilsberg (approved):
LGTM

PR #196, opened 2023-02-09
Block copolymer

jannisborn:
- Adding support for inference of models that use `PolymerGraphTokenizer` from RT package (at the moment only the `block copolymer` model)
- Adding support for training/finetuning such models
- All covered with unittests


src/gt4sd/domains/materials/__init__.py
```diff
@@ -33,6 +34,8 @@
 # TODO: setting to str directly requires no wrapping, so wrong strings could be passed
 Protein = str  # NewType('Protein', str)
 SMILES = str  # NewType('SMILES', str)
+SELFIES = str  # NewType('SELFIES', str)
+COPOLYMER = str  # NewType('COPOLYMER', str)
```
drugilsberg, line 38: Since it's not an abbreviation:
```suggestion
Copolymer = str  # NewType('Copolymer', str)
```

src/gt4sd/domains/materials/__init__.py
```diff
@@ -41,7 +44,13 @@
 Sequence = str
 
 
-def validate_molecules(
+class InputType(str, Enum):
+    selfies = "SELFIES"
+    smiles = "SMILES"
+    copolymer = "COPOLYMER"
```
drugilsberg, line 50: Being even more explicit would be good:
```suggestion
class MoleculeFormat(str, Enum):
    selfies = "SELFIES"
    smiles = "SMILES"
    copolymer = "Copolymer"
```

src/gt4sd/domains/materials/__init__.py
```diff
@@ -65,11 +74,70 @@ def validate_molecules(
...
+    """
+    # selfies is not valid if it contains -1 so return no valid ids
+    valid_ids = [i for i, s in enumerate(selfies_list) if s != -1]
+    selfies = [s for i, s in enumerate(selfies_list) if i in valid_ids]
+    return selfies, valid_ids
+
+
+def validate_copolymer(
+    copolymers_list: List[COPOLYMER],
+) -> Tuple[List[Chem.rdchem.Mol], List[int]]:
```
drugilsberg, line 96: Is typing right here? Probably it is not a `Chem.rdchem.Mol`

drugilsberg (commented):
Only a few suggestions on naming conventions.

PR #203, opened 2023-02-21
chore: improved error handling

jannisborn:


drugilsberg (commented):
Great and fast job.

What I would consider adding is a regex to post process the output of the sampling wrapper to replace the second occurrence of a property token (maybe getting them from the inference.json of the model)

drugilsberg (approved):
LGTM

PR #204, opened 2023-03-03
Enable Molformer into inference pipelines

christofid:
This PR introduces the following additions:
- Enable the use of finetuned MolFormer's variants via the inference pipelines
- Add notebook with MolFormer examples


drugilsberg (commented):
Looks good let's see what happens once the new package version is available.

drugilsberg (approved):

PR #205, opened 2023-03-08
MoLeR support for latent encoding

jannisborn:
- Enables previously unsupported feature in MoLeR: Initializing the decoder with latent codes from provided molecules. Thus far, the decoder was always initialized with randomly sampled latent codes.

- Expanded moler tests
- minor cleanups


drugilsberg (approved):
LGTM, great job,  we can put it in the queue for the merge

PR #210, opened 2023-03-22
chore: updating GPU build.

drugilsberg:
Minor improvements in RT training pipeline.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,9 +9,9 @@ joblib>=1.1.0
 gt4sd-molformer>=0.1.1
 keras>=2.3.1,<2.11.0
 keybert>=0.7.0
+markdown-it-py<3.0.0,>=2.2.0
 minio==7.0.1
 modlamp>=4.0.0
-molecule-generation>=0.3.0
 molgx>=0.22.0a1
 nglview>=3.0.3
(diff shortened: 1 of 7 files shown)


requirements.txt
```diff
@@ -9,9 +9,9 @@ joblib>=1.1.0
 gt4sd-molformer>=0.1.1
 keras>=2.3.1,<2.11.0
 keybert>=0.7.0
+markdown-it-py<3.0.0,>=2.2.0
```
jannisborn, line 12: If this is a requirement of the core package it should be added to the `setup.cfg` too, in order to keep things in sync as much as possible

jannisborn (commented):
Looks good, did you verify the new installation also locally (on Mac)?

requirements.txt
drugilsberg: it's not a direct requirement of core hence the decision to leave it out, it's a dependency that gets installed with a messy bound in conda builds.

jannisborn:
Since the installation for CPU also changed mildly, I think we should verify that it works on Mac. I'm doing this now, will approve the PR once my local tests pass

jannisborn:
tests pass locally

jannisborn (approved):

PR #211, opened 2023-04-11
feat: improving tokenizer handling in RT inference.

drugilsberg:


jannisborn (approved):
Good stuff! Great that we could resolve a few more shaky things in the RT inference. The sampling wrapper was build when we only had SELFIES models, so thx for fixing

PR #215, opened 2023-04-19
Use gt4sd-trainer-hf-pl training pipelines

christofid:


drugilsberg (commented):
Looks great, if ci passes we can merge

PR #217, opened 2023-05-04
feat: support custom (de)normalization of property values in RT

jannisborn:
Many RT models were trained on property values that were cast from the original range to a [0,1] range. Hence, the input and output property values, which should be in the original, user-understandable scale, have to be transformed before model execution.

Thus far, the only supported functions for (de)normalization was a linear transformation from the original scale to [0,1] and back.

This PR extends this functionality by allowing arbitrarily complex (de)normalization functions, in particular log-transformations. These functions are given as string lambda expressions in the model artifacts


drugilsberg (approved):
LGTM

PR #226, opened 2023-08-24
Fix paccmann gp stochasticity

### LINKED ISSUE #224: Bug with random seed in paccman generator
**Describe the bug**
PaccmannGP does not have deterministic generator behaviour although there is a code that sets a seed here
https://github.com/GT4SD/gt4sd-core/blob/ea9c299cea6ea6c63696b4e9ee49ceb16fb58507/src/gt4sd/algorithms/controlled_sampling/paccmann_gp/core.py#L191
it seems that this seed is only used for the optimizer but not for other sources of randomness in the code
https://github.com/GT4SD/gt4sd-core/blob/ea9c299cea6ea6c63696b4e9ee49ceb16fb58507/src/gt4sd/algorithms/controlled_sampling/paccmann_gp/implementation.py#L237C36-L237C36

I also tried the following after some googling without success,
it seems the seed should also be transferred to some network layer operation which I can't change when the model is fixed.
import numpy as np
import random
import tensorflow as tf

def seed_everything(seed=1234):
    random.seed(seed)
    os.environ['PYTHONHASHSEED'] = str(seed)
    np.random.seed(seed)
    tf.random.set_seed(seed)

def set_global_determinism(seed=1234):
    seed_everything(seed=seed)
    os.environ['TF_DETERMINISTIC_OPS'] = '1'
    os.environ['TF_CUDNN_DETERMINISTIC'] = '1'
    tf.config.threading.set_inter_op_parallelism_threads(1)
    tf.config.threading.set_intra_op_parallelism_threads(1)

jannisborn:
Closes #224 
PaccMannGP is now fully deterministic when a seed is given. Tests expanded to verify this.

Other:
- cleaning up the unused `generate_valid` methods in PaccMannRL
- fixing the pypi usage stats badges


drugilsberg (approved):
Lgtm

PR #232, opened 2023-11-22
Cleaned Enzeptional

yvesnana:
- Cleaned Enzeptional
- Refactored both Processing and Core files


drugilsberg (changes requested):
We are getting there. Address all remaining comments please, it should be straightforward. The CI is failing on black, ensure the styling is applied before commits.

PR #237, opened 2024-04-24
fix: fixed enzyme optimization with Kcat fitness function

yvesnana:
Resolved inconsistencies in core.py and updated the example file to ensure accurate execution of optimization with the KCAT fitness model.


examples/enzeptional/example_enzeptional.py
```diff
@@ -5,12 +5,16 @@
 from gt4sd.configuration import GT4SDConfiguration, sync_algorithm_with_s3
 
 
-def initialize_environment():
+def initialize_environment(model = "feasibility"):
```
drugilsberg, line 8: typing and docstring missing, please add

drugilsberg (changes requested):
Fix the build and address comments and then it's good to go.

PR #239, opened 2024-04-24
fix: fixed enzyme optmization with Kcat fitness function

yvesnana:


yvesnana:
Resolved inconsistencies in core.py and updated the example file to ensure accurate execution of optimization with the KCAT fitness model.

PR #241, opened 2024-05-23
Python310 support

jannisborn:
Also ramping up `pip` and some other dependencies.

Disabling clone protection due to this recent git LFS issue: https://github.com/git-lfs/git-lfs/issues/5749


jannisborn:
Changed base branch to facilitate commits by others

PR #247, opened 2024-06-10
Bumping pydantic >= 2.0.0

jannisborn:
Problem:
- HF Spaces apps are getting down due to deprecated gradio version
- Compliant gradio version relies on pydantic 2.*
- GT4SD relies on pydantic 1.*

--> Goal: Bump pydantic

Refactor is largely done but not all tests are passing yet.

Without the hacky line
```py
            VanillaConfiguration.__pydantic_validator__ = True  # type: ignore
```

we get:
```sh
TypeError: Parameter list to Generic[...] cannot be empty
```

It would be best to find a way to have a non-empty parameter list here.

### DIFF
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -14,3 +14,4 @@ sphinx>=5
 sphinx-autodoc-typehints==1.11.1
 jinja2<3.1.0
 sphinx_rtd_theme==0.5.1
+pydantic-settings>=2.0.0
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -13,5 +13,3 @@ pytest-cov==2.10.1
 sphinx>=5
 sphinx-autodoc-typehints==1.11.1
 jinja2<3.1.0
-sphinx_rtd_theme==0.5.1
-pydantic-settings>=2.0.0
--- a/src/gt4sd/configuration.py
+++ b/src/gt4sd/configuration.py
@@ -28,9 +28,8 @@ import os
 from functools import lru_cache
 from typing import Dict, Optional, Set
 
-from pydantic import BaseSettings
-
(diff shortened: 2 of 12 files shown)


dev_requirements.txt
```diff
@@ -13,4 +13,5 @@ pytest-cov==2.10.1
 sphinx>=5
 sphinx-autodoc-typehints==1.11.1
 jinja2<3.1.0
-sphinx_rtd_theme==0.5.1
\ No newline at end of file
+sphinx_rtd_theme==0.5.1
+pydantic-settings>=2.0.0
```
drugilsberg, line 17: this should be moved to the actual requirements as it's needed in the package at runtime

src/gt4sd/configuration.py
```diff
@@ -28,9 +28,8 @@
 from functools import lru_cache
 from typing import Dict, Optional, Set
 
-from pydantic import BaseSettings
-
 from .s3 import GT4SDS3Client, S3SyncError, sync_folder_with_s3, upload_file_to_s3
+from pydantic_settings import BaseSettings, SettingsConfigDict
```
drugilsberg, line 32: is isort fine with this?

drugilsberg (changes requested):
Great work, few comments.

src/gt4sd/configuration.py
fiskrt: Is isort still used? Seems to be commented out in `dev_requirements.txt`?

jannisborn: We removed it in #218 @drugilsberg. 
Still good practice to have absolute imports ahead of relative ones @fiskrt

drugilsberg: thanks for the inputs, agree with Jannis

fiskrt:
Can we merge?

drugilsberg (approved):
LGTM

PR #248, opened 2024-06-13
Pypi python version

jannisborn:


jannisborn:
Necessary as detailed here: https://github.com/actions/setup-python/issues/160

jannisborn:
To prevent builds from failing like this https://github.com/GT4SD/gt4sd-core/actions/runs/9498373751/job/26176967697

drugilsberg (approved):

PR #256, opened 2025-02-17
chore: update S3 keys in configuration.py

drugilsberg:


src/gt4sd/configuration.py
```diff
@@ -52,8 +52,8 @@ class GT4SDConfiguration(BaseSettings):
     gt4sd_disable_cudnn: bool = False
 
     gt4sd_s3_host: str = "s3.par01.cloud-object-storage.appdomain.cloud"
-    gt4sd_s3_access_key: str = "6e9891531d724da89997575a65f4592e"
-    gt4sd_s3_secret_key: str = "5997d63c4002cc04e13c03dc0c2db9dae751293dab106ac5"
+    gt4sd_s3_access_key: str = "b087e6810a5d4246a64e07e36ace338f"
+    gt4sd_s3_secret_key: str = "ba4a1db5647a32c6109b58714befb7ea7145b983143e0836"
```
jannisborn, line 56: I presume we also have to update the keys for the hub, right below in the code

jannisborn (changes requested):

src/gt4sd/configuration.py
drugilsberg: if I understood correctly the issue reported, those do not require the update.

drugilsberg: update as well after verifying expiration

jannisborn (approved):
Great we still have a functional version 👍🏼

